PR #1, opened 2020-12-12
Birmingham Class 2 - Chizim Chinuru - JS-Core-1 week 2

scarabeo7:
# Your Details

- Your Name: Chizim Chinuru
- Your City: Birmingham
- Your Slack Name: Chizim

# Homework Details

- Module: JavaScript Core 1
- Week: 2

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -10,7 +10,7 @@ var mentorCount = 9;
 var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom = roomMaxCapacity <= 25; // finish this statement
+var enoughSpaceInRoom = studentCount + mentorCount <= roomMaxCapacity; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity <= 25; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA === personB ; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
@@ -24,7 +24,7 @@ console.log(
   "Is there enough space in the room for all students and mentors?",
   enoughSpaceInRoom
 );
-console.log("Do person A and person B have the the same name?", sameName);
+console.log("Do person A and person B have the same name?", sameName);
 
 /* 
   EXPECTED RESULT
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -7,10 +7,12 @@
 */
 
 var name = "Daniel";
-var danielsRole = "Student";
+var danielsRole = "student";
 
-if(danielsRole === danielsRole){
-  console.log(`Hi, I'm ${name}, I'm a ${danielsRole}.`);
+if(danielsRole === "mentor"){
+  console.log(`Hi, I'm ${name}, I'm a mentor.`);
+}else{
+  console.log(`Hi, I'm ${name}, I'm a student.`);
 }
 
 /* 
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -7,7 +7,11 @@
 */
 
 var name = "Daniel";
-var danielsRole = "mentor";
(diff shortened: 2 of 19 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity <= 25; // finish this statement
```
ARTaylor1502, line 13: You need to take into consideration the number of students and mentors with this statement

exercises/E-conditionals/exercise.js
```diff
@@ -7,7 +7,11 @@
 */
 
 var name = "Daniel";
-var danielsRole = "mentor";
+var danielsRole = "Student";
+
+if(danielsRole === danielsRole){
```
ARTaylor1502, line 12: Theres no need for this conditional. It will always evaluate to true

exercises/F-logical-operators/exercise2.js
```diff
@@ -5,13 +5,39 @@
...
+
+function isShortName(name){
+  if (name.length <= 6){
+    return true;
+  }
+  return false;
+}
+
+function startsWithD(name){
+  if(name[0] === "D"){
```
ARTaylor1502, line 30: This is incorrect as it assumes the name parameter is an array and checks the first index. The parameter should be a string and so name.charAt(0)  === "D" would be more appropriate.

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[arr.length -1]; // complete this statement
```
ARTaylor1502, line 12: 👍

exercises/J-array-get-set/exercises2.js
```diff
@@ -7,6 +7,8 @@
 */
 
 var numbers = [1, 2, 3]; // Don't change this array literal declaration
+numbers[3] = 4; // One way of doing it. 
+numbers.push(5, 6); // Another way using .push() method which adds item to last of array
```
ARTaylor1502, line 11: Excellent 👌

mandatory/1-fix-functions.js
```diff
@@ -21,32 +19,24 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
-
-  if (isBigEnough) {
+  if (num > 10) {
```
ARTaylor1502, line 22: This is correct but I think the task wanted the isBigEough variable to be assigned the comparison expression result.

ie. let isBigEnough = num > 10;

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,47 @@
   Answer the following questions:
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
-  2. Remove the line that throws the error.
+    [ console.log(b) throws error as b is undefined]
+  2. Remove the line that throws the error. 
+    [commented out the line causing errors]
   3. What is printed to the console?
+    [2, 6, 6, 4, 7, 9, 8, 6, 9, 13, 10, 8]
   4. How many times is "f1" called?
```
ARTaylor1502, line 15: Its called twice. There are only 5 iterations of the for loop in total. On each loop the conditional checks if the current index is odd or even. The f1 is called only if the index is odd and as the index is only odd twice (1 and 3) the f1 function is only called twice.

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,47 @@
...
+    [10 times, used console.log("find", a)]
   5. How many times is "f2" called?
+    [2 times, used console.log("search", x)]
```
ARTaylor1502, line 18: It's called 3 times. As mentioned above the conditional checks if the index is odd or even. The index is even 3 times (0,2,4) and so f2 is called 3 times.

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,47 @@
...
   6. What value does the "a" parameter take in the first "f1" call?
+    [value a = 9, used console.log("before", a)]
```
ARTaylor1502, line 20: I think you've slightly misunderstood the question. It's asking for the 'a' parameter. As in the first parameter of the f1 function. This is passed in as the variable 'i' when the f1 function is called in the for loop. As the f1 function is called when the index is first an odd number on its second iteration of the loop, the index is incremented once to make it 1.

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,47 @@
...
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
+    [value a = 10, used console.log("after", a)]
```
ARTaylor1502, line 22: The value of 'a' is 8. The value of 'a' is initially 6. The f1 function first gets called on the second iteration of the for loop in which time the 'a' variable has incremented by 1 twice. to give a value of 8.

ARTaylor1502:
Very Good 👍

PR #10, opened 2020-12-16
West Midlands 2 - Jasvir Kaur - JavaScript Core 1 - week 2

JK1320:
# Your Details

- Your Name: Jasvir Kaur
- Your City: Walsall
- Your Slack Name: Jasvir Kaur

# Homework Details

- Module: JavaScript Core 1
- Week: 2

### DIFF
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -9,6 +9,10 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+if (true) {
+console.log("Hi, I'm " + name + "," + " I'm a " + danielsRole + ".")
+}
+
 /* 
 EXPECTED RESULT
 ---------------
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -10,7 +10,9 @@ var name = "Daniel";
 var danielsRole = "mentor";
 
 if (true) {
-console.log("Hi, I'm " + name + "," + " I'm a " + danielsRole + ".")
+console.log(`Hi, I'm ${name}, I'm a ${danielsRole}.`);
+} else {
+  console.log(`Hi, I'm ${name}, I'm a student`);
 }
 
 /* 
--- a/exercises/F-logical-operators/exercise2.js
+++ b/exercises/F-logical-operators/exercise2.js
@@ -20,14 +20,14 @@ function isBetween5and10(range){
 }
 
 function isShortName (name){
-  if (name = "john" || name < "john"){
+  if (name.length < 10){
     return true;
   }
   return false;
 }
 
 function startsWithD(word){
-  if (word === "Daniel") {
+  if (word.startsWith("D")) {
     return true;
   }
   return false;
--- a/exercises/F-logical-operators/exercise2.js
+++ b/exercises/F-logical-operators/exercise2.js
@@ -5,7 +5,33 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(number) {
+if (number < 0) {
+  return true;
+}
+return false;
+}
+
+function isBetween5and10(range){
+  if (range > 5 && range <= 10){
+    return true;
+  }
+  return false;
+}
+
+function isShortName (name){
+  if (name = "john" || name < "john"){
+    return true;
+  }
+  return false;
+}
+
+function startsWithD(word){
+  if (word === "Daniel") {
+    return true;
+  }
+  return false;
+}
 
 /* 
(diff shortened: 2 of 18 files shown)


exercises/E-conditionals/exercise.js
```diff
@@ -9,6 +9,12 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+if (true) {
```
ARTaylor1502, line 12: This doesn't make sense. True will always evaluate true and so the else statement will never be reached. I would instead do something like:

if(danielsRole === "mentor")

exercises/F-logical-operators/exercise2.js
```diff
@@ -5,7 +5,33 @@
...
-function isNegative() {}
+function isNegative(number) {
+if (number < 0) {
+  return true;
+}
+return false;
+}
+
+function isBetween5and10(range){
+  if (range > 5 && range <= 10){
```
ARTaylor1502, line 16: Should be >= 5

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,12 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+  if (sentence.includes("code")) {
+    return "true";
+  } else
```
ARTaylor1502, line 14: Careful with formatting. Should wrap else statement in curly braces.

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+ return arr[arr.length - 1]
```
ARTaylor1502, line 12: Nice 👍

extra/1-radio-stations.js
```diff
@@ -26,6 +52,12 @@
  */
 // `getStations` goes here
 
+function getStations (){
+  const frequenciesAvailable = getAllFrequencies();
+  const radioFrequencies = frequenciesAvailable.filter(frequencies => isRadioStation(frequencies));
+  return radioFrequencies;
```
ARTaylor1502, line 58: Would probably return the expression directly

i.e  return frequenciesAvailable.filter(frequencies => isRadioStation(frequencies));

mandatory/1-fix-functions.js
```diff
@@ -21,30 +20,30 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = 10;
 
-  if (isBigEnough) {
+  if (isBigEnough < num ) {
```
ARTaylor1502, line 25: Think this should be a conditional statement:

let isBigEnough = num > 10;

mandatory/2-function-creation.js
```diff
@@ -34,7 +54,16 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(arr) {}
+function formatPercentage(arr) {
+  for (let i = 0; i < arr.length; i++) {
+if (arr[i] > 100) {
+  arr[i] = 100;
+}
+ arr[i] = Math.round(arr[i] * 100) / 100 +"%";
```
ARTaylor1502, line 62: Careful with your tabbing. Try to tab when entering functions and loops

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,19 @@
...
   2. Remove the line that throws the error.
+      commented out console.log(b);
   3. What is printed to the console?
+    2, 6, 4, 9, 6, 13, 8
   4. How many times is "f1" called?
+      2
   5. How many times is "f2" called?
+      3
   6. What value does the "a" parameter take in the first "f1" call?
+      7
```
ARTaylor1502, line 20: This is actually 1. The 'a' parameter is passed in as the 'i' variable in the f1 on the second iteration of the for loop

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,19 @@
...
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
+      6
```
ARTaylor1502, line 22: This is 8. The initial value of 'a' is 6 and incremented by 1 with each iteration of the for loop. As the f1 loop is only called on the second iteration as the 'i' variable is odd at that point the 'variable' is incremented by 1 twice

ARTaylor1502:
Very good 👍

PR #23, opened 2020-12-20
completed my assignment

Sozchi:
# Your Details

- Your Name: Ozioma Ukaonu
- Your City: west Midlands
- Your Slack Name: Sozchi

# Homework Details

- Module: JavascriptCore1
- Week:week 3

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = true; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = true; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = false; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/D-predicates/exercise.js
+++ b/exercises/D-predicates/exercise.js
@@ -6,10 +6,16 @@
 */
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative( number ){
+   return typeof number === "number";
+}
+  isNegative(5)
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10( number ){
+    return typeof number === "number";
+}
+   isBetweenZeroAnd10(10)
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -8,6 +8,13 @@
 
 var name = "Daniel";
 var danielsRole = "mentor";
+if ( danielsRole )
+{
+  console.log( "Hi, I'm Daniel, I'm a mentor." );
+} else
+{
+  console.log( "Hi, I'm Daniel, I'm a student." );
+}
 
 /* 
 EXPECTED RESULT
--- a/exercises/F-logical-operators/exercise.js
+++ b/exercises/F-logical-operators/exercise.js
@@ -9,16 +9,22 @@
 var htmlLevel = 8;
 var cssLevel = 4;
 
+
+
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
-var htmlLevelAbove5;
-var cssLevelAbove5;
+var htmlLevelAbove5 = htmlLevel > 5;
+var cssLevelAbove5 = cssLevel > 5;
+
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-var cssAndHtmlAbove5;
-var cssOrHtmlAbove5;
+var cssAndHtmlAbove5 = htmlLevel > 5 && cssLevel > 5;
+var cssOrHtmlAbove5 = htmlLevel > 5 || cssLevel > 5;
+
+
+
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/F-logical-operators/exercise2.js
+++ b/exercises/F-logical-operators/exercise2.js
@@ -5,8 +5,21 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative( number ){
+  return number < 0;
+}
 
+function isBetween5and10( number){
+  return number >= 5 && number <= 10;
+}
+
+function isShortName(name){
+  return name.length < 7;
+}
+
+function startsWithD(name){
+  return name.charAt( 0 ) === "D";
+}
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
--- a/exercises/G-conditionals-2/exercise-1.js
+++ b/exercises/G-conditionals-2/exercise-1.js
@@ -6,7 +6,13 @@
   - if number is more or equal to zero, return the word "positive"
 */
 
-function negativeOrPositive(number) {}
+function negativeOrPositive( number ){
+  if ( number >= 0 ){
+    return "positive";
+  } else { 
+    return "negative";
+  }
+} 
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-2.js
+++ b/exercises/G-conditionals-2/exercise-2.js
@@ -7,7 +7,15 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed( grade ){
+  if ( grade <= 49 )
+  {
+    return "failed";
+  } else if ( grade >= 50 )
+  {
+    return "passed";
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-3.js
+++ b/exercises/G-conditionals-2/exercise-3.js
@@ -8,7 +8,16 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade( mark ){
+  if ( mark <= 49){
+    return "F";
+  } else if ( mark >= 80 ){
+    return "A";
+  } else if ( mark > 60 ){
+    return "B";
+  } else ( mark <= 50 )
+    return "C";
+} 
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-4.js
+++ b/exercises/G-conditionals-2/exercise-4.js
@@ -8,7 +8,13 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode( sentence ){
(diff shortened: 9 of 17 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
...
+var moreStudentsThanMentors = true; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = true; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = false; // finish this statement
```
ARTaylor1502, line 17: The task required you to use comparison operators to evaluate the boolean value of the variables. You've directed assigned the boolean value.

i.e The first one would be

var moreStudentsThanMentors = studentCount > mentorCount;

exercises/D-predicates/exercise.js
```diff
@@ -6,10 +6,16 @@
...
+   return typeof number === "number";
+}
+  isNegative(5)
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10( number ){
+    return typeof number === "number";
+}
+   isBetweenZeroAnd10(10)
```
ARTaylor1502, line 18: These two function are just checking the type of the parameter being passed in to the function rather than following the functionality inferred by the function name.

exercises/E-conditionals/exercise.js
```diff
@@ -8,6 +8,13 @@
 
 var name = "Daniel";
 var danielsRole = "mentor";
+if ( danielsRole )
```
ARTaylor1502, line 11: This should check specifically for the mentor role. You're effectively just checking if the string is a non null value which will always evaluate to true.

It should be:

if(danielsRole == 'mentor')

exercises/F-logical-operators/exercise.js
```diff
@@ -9,16 +9,22 @@
...
+var cssLevelAbove5 = cssLevel > 5;
+
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-var cssAndHtmlAbove5;
-var cssOrHtmlAbove5;
+var cssAndHtmlAbove5 = htmlLevel > 5 && cssLevel > 5;
+var cssOrHtmlAbove5 = htmlLevel > 5 || cssLevel > 5;
```
ARTaylor1502, line 24: Excellent 👍

exercises/F-logical-operators/exercise2.js
```diff
@@ -5,8 +5,21 @@
...
+  return number >= 5 && number <= 10;
+}
+
+function isShortName(name){
+  return name.length < 7;
+}
+
+function startsWithD(name){
+  return name.charAt( 0 ) === "D";
+}
```
ARTaylor1502, line 22: 👌

exercises/G-conditionals-2/exercise-1.js
```diff
@@ -6,7 +6,13 @@
...
 */
 
-function negativeOrPositive(number) {}
+function negativeOrPositive( number ){
+  if ( number >= 0 ){
+    return "positive";
+  } else { 
+    return "negative";
+  }
+}
```
ARTaylor1502, line 15: 👍

exercises/G-conditionals-2/exercise-2.js
```diff
@@ -7,7 +7,15 @@
...
-function studentPassed(grade) {}
+function studentPassed( grade ){
+  if ( grade <= 49 )
+  {
+    return "failed";
+  } else if ( grade >= 50 )
+  {
+    return "passed";
+  }
+}
```
ARTaylor1502, line 18: 👍

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,16 @@
...
+function calculateGrade( mark ){
+  if ( mark <= 49){
+    return "F";
+  } else if ( mark >= 80 ){
+    return "A";
+  } else if ( mark > 60 ){
+    return "B";
+  } else ( mark <= 50 )
+    return "C";
+}
```
ARTaylor1502, line 20: Excellent 👍

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,13 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode( sentence ){
+  if ( sentence === sentence1  ){
```
ARTaylor1502, line 12: This comparison is incorrect you should use a function to check if the word 'code' is a substring of the string passed in as a parameter. You could use the 'includes' function to achieve this for exampe:

https://flaviocopes.com/how-to-string-contains-substring-javascript/

exercises/H-array-literals/exercise.js
```diff
@@ -4,8 +4,8 @@
   Declare some variables assigned to arrays of values
 */
 
-var numbers = []; // add numbers from 1 to 10 into this array
-var mentors; // Create an array with the names of the mentors: Daniel, Irina and Rares
+var numbers = [1,2,3,4,5,6,7,8,9,10]; // add numbers from 1 to 10 into this array
+var mentors = ['Daniel', 'Irina', 'Rares']; // Create an array with the names of the mentors: Daniel, Irina and Rares
```
ARTaylor1502, line 8: 👍

exercises/I-array-properties/exercise.js
```diff
@@ -6,7 +6,7 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  return ( arr === names); // complete this statement
```
ARTaylor1502, line 9: This doesn't make sense. You're comparing the array passed as a parameter to a global names array. 

You should just be checking the array passed in to the function's length:

i.e return array.length === 0;

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,13 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return numbers[0]; {
```
ARTaylor1502, line 8: The first return value is correct as you are returning the first index of the array.The second return statement should be removed

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,13 @@
...
 function first(arr) {
-  return; // complete this statement
+  return numbers[0]; { 
+    return numbers[2]
+  }// complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return names[3]; // complete this statement
```
ARTaylor1502, line 14: Whilst this is technically correct, this method isn't dynamic or scalable. This function will always return the third index of an array regardless of the size of the array passed into it. A better and more dynamic way to do this would be to calculate the last index of the array passed into the function like this:

return names[names.length - 1]

This will count the amount of items in the array. The -1 will then take into account the fact array use a zero based index and calculate the correct array index for you.

exercises/J-array-get-set/exercises2.js
```diff
@@ -7,7 +7,7 @@
 */
 
 var numbers = [1, 2, 3]; // Don't change this array literal declaration
-
+let newNumbers = numbers.push(4)
```
ARTaylor1502, line 10: 👌

mandatory/1-fix-functions.js
```diff
@@ -9,42 +9,40 @@
...
   } else {
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
 
-  if (isBigEnough) {
+  if (num > 10 ) {
```
ARTaylor1502, line 24: I would have personally done

let isBigEnough = num > 10

What you've done is correct though

mandatory/2-function-creation.js
```diff
@@ -15,15 +28,25 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate( num ){
+  if ( typeof num === "number" && num % 2 === 0 && num <= 100 ) {
```
ARTaylor1502, line 32: To keep things more readable it might be better to use predicates assigned to variables.

i.e

const isNumber = typeof num === "number";
const isEven = num % 2 === 0
const isLessThanOrEqualTo100 = num <= 100

if(isNumber && isEven && isLessThanOrEqualTo100)

mandatory/2-function-creation.js
```diff
@@ -15,15 +28,25 @@ Complete the function to check if the variable `num` satisfies the following req
...
+}
+console.log( validate( 10 ) );
 /* 
 Write a function that returns a copy of the given array arr, but with the element at the given index, index removed.
 The function must NOT change the original array, arr.
 */
 
-function remove(arr, index) {
-  return; // complete this statement
+function remove( arr, index ){
```
ARTaylor1502, line 47: 👍

mandatory/2-function-creation.js
```diff
@@ -34,7 +57,18 @@ Write a function that:
...
 
-function formatPercentage(arr) {}
+function formatPercentage( arr ){
+  for ( i = 0; i < arr.length; i++ ) {
+    if ( arr[i] > 100 ) {
+      arr[i] = "100%";
+    }
+    else
+    {
+      arr[i] = Math.round( ( arr[i] + Number.EPSILON ) * 100 ) / 100 + "%";
```
ARTaylor1502, line 67: I think you would be ok here with just:

arr[i] = arr[i].toFixed(2) + "%";

mandatory/3-playing-computer.js
```diff
@@ -28,7 +28,7 @@ const f2 = function (a, b) {
 
 console.log(x);
 console.log(a);
-console.log(b);
+// console.log(b); // It throws an error because b is undefined
```
ARTaylor1502, line 31: Nice 👍

mandatory/3-playing-computer.js
```diff
@@ -8,11 +8,11 @@
...
   1. This program throws an error. Why? (If you can't find it, try executing it).
   2. Remove the line that throws the error.
-  3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  3. What is printed to the console? 2,6,4,9,6,18,8
+  4. How many times is "f1" called? 2
+  5. How many times is "f2" called? 2
```
ARTaylor1502, line 13: There are 5 iterations of the for loop and the index which starts at 0 is incremented by one with each loop until it reaches 5, then it stops. For each iteration a check is done on the current index. if it is even the f2 function is called. The index is 0,2 and 4 at various iterations of the loop and so the f2 function is called 3 times. The index is also 1 and 3 at certain iterations and so the f1 function is called twice.

mandatory/3-playing-computer.js
```diff
@@ -8,11 +8,11 @@
...
+  6. What value does the "a" parameter take in the first "f1" call? 6
```
ARTaylor1502, line 14: This is actually 0. The first call to f1 passes the 'i' variable as the a parameter. On the first iteration of the loop 'i' is 0.

mandatory/3-playing-computer.js
```diff
@@ -8,11 +8,11 @@
...
+  7. What is the value of the "a" outer variable when "f1" is called for the first time? 8
```
ARTaylor1502, line 15: Correct. Very good 🎉

PR #27, opened 2021-01-24
Cape Town, Rodrigue Ndzana,JavaScript-Core-1-Coursework-Week2

RodrigueNdzana:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -5,7 +5,15 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
+  for (let b=0; b < strArr.length; b++){
+    let removeSpace= strArr[b].trim(); // the Trim() method will remove whitespace from both ends of the string.
+    let removeForwardSlash=removeSpace.replace("/",''); // it will remove by replacing any forward slashes by new string
+    let lowercaseString=removeForwardSlash.toLowerCase();
+    strArr[b]=lowercaseString;
+  }
+  return strArr;
+}
 
 /*
 Complete the function to check if the variable `num` satisfies the following requirements:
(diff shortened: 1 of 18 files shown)


LawrenceBaatjies:
good job

ShawenHarker:
I really have to commend you on your commit and push to github, that's really well did on your end.

mandatory/2-function-creation.js
```diff
@@ -5,7 +5,15 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
```
Ntshembo-Hlongwane1, line 8: ### Hey there so I just have some code refactoring tips for you.

- So I am pretty sure your **tidyUpString function** does pass all test cases, however we can refactor it a bit
- So good thing you are already looping and indexing through each each element
- So I do not know if you know that in JS we can chain methods like this below

```js
.replace("/", "").trim().toLowerCase();
```

```diff
- let removeSpace = strArr[b].trim(); // the Trim() method will remove whitespace from both ends of the string.
- let removeForwardSlash = removeSpace.replace("/", '') // it will remove by replacing any forward slashes by new string
- let lowercaseString = removeForwardSlash.toLowerCase(); // makes the string all lowercase
+ strArr[b] = strArr[b].replace("/", "").trim().toLowerCase();
```
- Now all those 3 lines of code are just boiled down to just **1 single line**

### However we could still work on this code a bit more by making our lives easier on the looping side 

```js
let count = 0;
for (string of strArr){
    strArr[count] = string.replace("/", "").trim().toLowerCase();;
    count++;
};
```
[Refer to this link to check how that kind of for loop work's it is a game changer in certain cases](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of)

### Hope this was helpful 🔥💪

mulindierick: this is a good idea. I think the reason why he used the previous method is because of the scope of what we have learned so far. the "for... of" loop is in ES6 and I think most people might find it difficult to implement or don't know about it. 
I also used your approach, but with a map() instead of for of loop. 

thanks so, it's good that you keep bringing this up. It opens a door for further learning

PR #32, opened 2021-01-25
London Class 7 - Yunus FIRAT - Javascript Core - Week 2

yunusfirat:
# Your Details

- Your Name: Yunus FIRAT
- Your City: London
- Your Slack Name:@Yunus FIRAT

# Homework Details

- Module: JavaScript Core
- Week: 2

### DIFF
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -9,6 +9,12 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+if (danielsRole === "mentor") {
+  console.log("Hi, I'm Daniel, I'm a mentor.");
+} else {
+  console.log("Hi, I'm Daniel, I'm a student.");
+}
+
 /* 
 EXPECTED RESULT
 ---------------
--- a/exercises/J-array-get-set/exercises2.js
+++ b/exercises/J-array-get-set/exercises2.js
@@ -7,6 +7,7 @@
 */
 
 var numbers = [1, 2, 3]; // Don't change this array literal declaration
+numbers[3] = 4;
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -5,7 +5,12 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
+  for (i = 0; i < strArr.length; i++) {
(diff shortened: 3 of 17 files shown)


exercises/E-conditionals/exercise.js
```diff
@@ -9,6 +9,12 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+if (danielsRole === "mentor") {
+  console.log("Hi, I'm Daniel, I'm a mentor.");
+} else {
```
akbasmustafa, line 14: I works. my minor offer for best practice that "else if " can be preferred rather than "else" to code  'If Daniel is a student, ...' condition.

exercises/J-array-get-set/exercises2.js
```diff
@@ -7,6 +7,7 @@
 */
 
 var numbers = [1, 2, 3]; // Don't change this array literal declaration
+numbers[3] = 4;
```
lpalma, line 10: Hey Yunus :wave: 
What if we had to multiple elements to the end of the array? Would we have to pass the index each time?
Is there another way to insert an element at the end of the array?

mandatory/2-function-creation.js
```diff
@@ -5,7 +5,12 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
+  for (i = 0; i < strArr.length; i++) {
```
lpalma, line 9: Nice!
Could we rewrite this function using `map` or `forEach`? How would that look? :)

mandatory/2-function-creation.js
```diff
@@ -15,15 +20,23 @@ Complete the function to check if the variable `num` satisfies the following req
...
 Write a function that returns a copy of the given array arr, but with the element at the given index, index removed.
 The function must NOT change the original array, arr.
 */
 
 function remove(arr, index) {
-  return; // complete this statement
+  let newArr = [];
+  for (elem of arr) {
+    newArr.push(elem);
+  }
```
lpalma, line 37: Well done :+1: 
Is there a way to return a copy of an array without having to iterate through each element? ;)

mandatory/2-function-creation.js
```diff
@@ -34,7 +47,17 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(arr) {}
+function formatPercentage(arr) {
+  for (i = 0; i < arr.length; i++) {
```
lpalma, line 51: Nice :+1: 
How would this function look like if we rewrite it using `map`? ;)

PR #33, opened 2021-01-26
London Class 7 - Askin Ekinci - Javascript Core 1 - Week 2

askinekincihf:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = true;
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = true;
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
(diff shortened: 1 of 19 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors = true;
+var moreStudentsThanMentors = studentCount > mentorCount;
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom = true;
+var enoughSpaceInRoom = roomMaxCapacity > (studentCount > mentorCount);
```
ademalagoz, line 13: Hi Askin.Great work. I only suggest this part because it may not be working properly for all other codes. Maybe you can consider below alternatives. Expect for this you are amazing.
var enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount;
var enoughSpaceInRoom = roomMaxCapacity >= 25;

askinekincihf: Many thanks for the feedback Adem. Yes, you are right, I made a small mistake while considering the goal of the variable. I have just modified the variable named enoughSpaceInRoom.

PR #40, opened 2021-01-26
London Class 7- Yasemin Bayraktutan-JavaScript Core 1- Week2

ybayraktutan:
exercises comleted

# Your Details

- Your Name: Yasemin Bayraktutan 
- Your City: London
- Your Slack Name: Yasemin

# Homework Details

- Module:  JavaScript Core 1-
- Week: 2

### DIFF
--- a/exercises/D-predicates/exercise.js
+++ b/exercises/D-predicates/exercise.js
@@ -6,10 +6,14 @@
 */
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  return number < 0;
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  return number < 10 && number > 0;
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-3.js
+++ b/exercises/G-conditionals-2/exercise-3.js
@@ -11,9 +11,9 @@
 function calculateGrade(mark) {
   if(mark >= 80){
     return "A";
-  } else if (mark < 80 && mark > 60) {
+  } else if (mark > 60) {
     return "B";
-  } else if (mark <= 60 && mark >50) {
+  } else if (mark >=50) {
     return "C";
   } else {
     return "F";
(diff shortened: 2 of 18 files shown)


exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,17 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if(mark >= 80){
+    return "A";
+  } else if (mark < 80 && mark > 60) {
+    return "B";
+  } else if (mark <= 60 && mark >50) {
```
akbasmustafa, line 16: my explanation to '...no lower than 50...' that is should be cover "equal and greater than 50"

akbasmustafa (changes requested):

exercises/D-predicates/exercise.js
```diff
@@ -6,10 +6,14 @@
...
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  return number < 0;
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
```
AhmadJanah, line 14: Hi Yasemin,

You did great work, congratulation,

I think between should include the numbers of 0, and 10 as well, what do you think?

Regards,

exercises/G-conditionals-2/exercise-3.js
AhmadJanah, line 16: Your work here is definitely correct, but you can make it shorter by removing some code, for example you can remove  (mark < 80 && ) from second if, as the compiler will come here only when mark < 80. your code can be:
if(mark >= 80){
    return "A";
  } else if (mark > 60) {
    return "B";
  } else if (mark >50) {

AhmadJanah (commented):
Thank you Yasemin for this great work, I hope you will accept my comments leniency.
Your code is amazing, but I added some extra approvments.

Best Wishes,

exercises/G-conditionals-2/exercise-3.js
ybayraktutan: Thank you very much for your suggestion. I will amend my code.

ybayraktutan: Thank you very much for your suggestion. I will amend my code.

PR #43, opened 2021-01-28
London Class 7 - Negin Dn- JavaScript-Core-1-Week 2

NeginDn:
# Your Details

- Your Name: Negin Dn
- Your City: London
- Your Slack Name: Negin Dn

# Homework Details

- Module:
- Week:


exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,12 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if ((mark) >= 80) return "A";
+  else if (mark < 80 && mark > 60) return "B";
```
nPelvin, line 13: Hey Negin,
Hope you are well and having a good week.
I've reviewed your code - well done. I even found a few ways I could improve my own.

I just wanted to mention one way I noticed that you could shorten this function. 

If you look at line 13 you wouldn't need to check that the mark is smaller than 80 as it was already checked at line 12 before passing down to the else statement. Something like: **else if (mark > 60) return "B";** would be enough. You can also do something similar with line 14. 

Hope that makes sense. Let me know if any issues.

Speak soon.

Nigel

mandatory/2-function-creation.js
```diff
@@ -1,12 +1,18 @@
...
+/*[]
 Write a function that:
 - takes an array of strings as input
 - removes any spaces in the beginning or end of the strings
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
-
+function tidyUpString(strArr) {
```
askinekincihf, line 8: Hi Negin, you have done a really great work, well done. I have got only feeling regarding how to shorten this code from line 10 to 13. Your code really makes sense, however why don't we write all three lines in only one line in this particular exercise in order not to repeat our code.

mandatory/1-fix-functions.js
```diff
@@ -22,31 +21,27 @@ function mood() {
...
 }
 
 function sortArray(letters) {
-  let sortedLetters = letters;
-
+  let sortedLetters = letters.sort();
   return sortedLetters;
 }
 
 function first5(numbers) {
```
askinekincihf, line 36: let is assigned undefined first and only after one line it is reassigned. What do you think if we rewrite line 38 as `let sliced = numbers.slice(0, 5);` then delete the line 37?

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[arr.length - 1]; // complete this statement
```
askinekincihf, line 12: That is on spot. we can basically find the last element of array with your code.  well done.

exercises/F-logical-operators/exercise2.js
```diff
@@ -5,7 +5,27 @@
...
 
-function isNegative() {}
+function isNegative(number1) {
+  let result = number1 < 0;
+  return result;
+}
+
+function isBetween5and10(number1) {
+  let result = number1 <= 10 && number1 > 5;
+  return result;
```
ybayraktutan, line 15: You can shorten your code by using this "function isNegative(number) {return number < 0 ; } 
and it is same for others.

ybayraktutan (commented):
Great work Negin! Your code is amazing. I have suggested some minor improvements.
Have a nice weekend.

exercises/G-conditionals-2/exercise-3.js
NeginDn: Hi Nigel,

Thank you for your kind feedbacks. You are very right and this not only helped me have shorten codes, it also helped me understand the conditional logic a bit more clear, I'll change the code. Thank you :)

mandatory/2-function-creation.js
NeginDn: Hi Askin,

Thank you for the feedback. It completely makes sense, but I am not sure how to combine the three line together. I really appreciate if you can help me with it. Thank you :)

mandatory/1-fix-functions.js
NeginDn: Very true, Done. Thank you :)

exercises/F-logical-operators/exercise2.js
NeginDn: Hi Yasamin, Thank you for the feedbacks, You mean not to assign it to a variable result? If yes, you are right there isn't any need for the result variable just find this a bit more tidy. Please let me know if you meant something else. Thank you  :)

mandatory/2-function-creation.js
askinekincihf: I think the code below could be used to combine three lines into one line.
`strArr[i] = strArr[i].trim().replace("/", "").toLowerCase();`

exercises/E-conditionals/exercise.js
```diff
@@ -8,7 +8,11 @@
 
 var name = "Daniel";
 var danielsRole = "mentor";
-
+if ((danielsRole = "mentor")) {
```
louisechow, line 11: Careful here, you are using the `=` which is the operator to assign a value. I think you want to do a comparison here. What operator is used for comparison?

exercises/E-conditionals/exercise.js
```diff
@@ -8,7 +8,11 @@
...
+  console.log("Hi, I'm Daniel, I'm a mentor.");
+} else {
+  console.log("Hi, I'm Daniel, I'm a student.");
+}
 /*
```
louisechow, line 16: An extra challenge for yourself would be to change the code so that your console message returns the correct regardless of what role I assign to Daniel.

Right now, if I change `danielsRole = "teacher"` then it will print out `Hi, I'm Daniel, I'm a student`. What would you change?

exercises/F-logical-operators/exercise2.js
louisechow: In this case I would just use `return number1 < 0`.

Declaring variables in programming takes up some of your computer's memory when your program is run. In this scenario, the equation is simple enough that I don't think you need to assign it to a variable for clarity first.

exercises/F-logical-operators/exercise2.js
```diff
@@ -5,7 +5,27 @@
...
 */
 
-function isNegative() {}
+function isNegative(number1) {
+  let result = number1 < 0;
+  return result;
+}
+
+function isBetween5and10(number1) {
+  let result = number1 <= 10 && number1 > 5;
```
louisechow, line 14: I think this function should only be true for 6, 7, 8, 9. 
But right now it will return true for 10.

exercises/F-logical-operators/exercise2.js
```diff
@@ -5,7 +5,27 @@
...
+  let result = name1.length;
+  if (result < 10) return true;
+  else return false;
+}
+
+function startsWithD(name1) {
+  if (name1.charAt(0) === "D") {
+    return true;
+  } else return false;
+}
```
louisechow, line 28: I would suggest for `if... else....` blocks of code that you continue to use the curly braces and indentation even when you only have one statement inside the braces. This makes it very clear what gets run in each condition and it's easier to spot mistakes. 

E.g.
```
if (result < 10) {
    return true;
} else {
    return false;
}
```
It is accepted to miss out the braces if you have only one statement being run, but in that case you should add indentation.
E.g.
```
if (result < 10)
    return true;
else
    return false;
```

exercises/G-conditionals-2/exercise-1.js
```diff
@@ -6,9 +6,12 @@
   - if number is more or equal to zero, return the word "positive"
 */
 
-function negativeOrPositive(number) {}
+function negativeOrPositive(number) {
```
louisechow, line 9: You could get away with just an `if...else...` here instead of `if... else if....`. A number can only be either positive or negative.

exercises/G-conditionals-2/exercise-1.js
```diff
@@ -6,9 +6,12 @@
...
+  if (number < 0) return "negative";
+  else if ((number) => 0) return "positive";
+}
```
louisechow, line 12: See the comment I left in `F-logical-operators/exercise2.js` about the best way to format the code.

exercises/G-conditionals-2/exercise-2.js
```diff
@@ -7,7 +7,10 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if (grade < 50) return "failed";
+  else if (grade => 50) return "passed"
+}
```
louisechow, line 13: Similar to exercise1 you could just use if...else...

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,12 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80) return "A";
+  else if (mark > 60) return "B";
+  else if (mark > 50) return "C";
```
louisechow, line 14: I think that a mark of 50 should give you a "C".

mandatory/1-fix-functions.js
```diff
@@ -11,8 +11,7 @@
 */
 
 function mood() {
```
louisechow, line 13: I think I just found a mistake in the homework 😅 , Line 13 should read `function mood(isHappy)`.

The answer is to remove line 14 as you have done, line 15 shouldn't be necessary. Sorry about the mistake in the homework.

mandatory/1-fix-functions.js
```diff
@@ -22,31 +21,26 @@ function mood() {
 
 function greaterThan10(num) {
```
louisechow, line 22: The argument `num` that is passed into your function is the number that you are checking. You should define isBigEnough to indicate if that number is greater than 10.

mandatory/2-function-creation.js
louisechow: Askin has made a great suggestion here. This is called `Function chaining/Method chaining`. 
https://dev.to/nedsoft/method-chaining-in-javascript-3klb

mandatory/2-function-creation.js
```diff
@@ -34,7 +47,14 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(arr) {}
+function formatPercentage(arr) {
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i] > 100) arr[i] = 100;
+    arr[i] = +arr[i].toFixed(2);
```
louisechow, line 53: I had to go look up what the unary operator did it because I wasn't familiar with it 😅 .

I'd say that you'd see `parseFloat()` used more commonly, but it's fine for you to use the unary operator instead (as long as you understand what it's doing).

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,17 @@
...
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
+     b has never been defined.
   2. Remove the line that throws the error.
+     console.log(b); has been removed.
   3. What is printed to the console?
+    2, 6, 4, 9, 6, 13, 8 
   4. How many times is "f1" called?
+    Three
   5. How many times is "f2" called?
```
louisechow, line 17: `f1` gets called each time `i` is an odd number, so I don't think 3 is the correct answer.

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,17 @@
...
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+    Two
```
louisechow, line 18: See above comment

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,17 @@
   Answer the following questions:
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
+     b has never been defined.
```
louisechow, line 10: ✅

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,17 @@
...
   2. Remove the line that throws the error.
+     console.log(b); has been removed.
   3. What is printed to the console?
+    2, 6, 4, 9, 6, 13, 8
```
louisechow, line 14: ✅

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,17 @@
...
+     console.log(b); has been removed.
   3. What is printed to the console?
+    2, 6, 4, 9, 6, 13, 8 
   4. How many times is "f1" called?
+    Three
   5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+    Two
+  6. What value does the "a" parameter take in the first "f1" call? 8
```
louisechow, line 19: ✅

mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,17 @@
...
+  7. What is the value of the "a" outer variable when "f1" is called for the first time? 8
```
louisechow, line 20: I think here, by outer variable we mean the one on line 19

louisechow:
Hi Negin, you did a great job with the homework 🌟 

One thing to pay attention to is how you are formatting your `if...else...` blocks. I made a suggestion of a format for you to stick to which I think is much easier to read.

There were a few exercises that I left comments on for you to revisit. Let me know if you have any questions.

mandatory/2-function-creation.js
NeginDn: Thank you, I've changed it

exercises/F-logical-operators/exercise2.js
NeginDn: Great to know, Thank you both, I've change it

exercises/F-logical-operators/exercise2.js
NeginDn: Oh I see. I thought it looks more professional this way :D,. Obviously it's not, I'll change the code, thank you

exercises/E-conditionals/exercise.js
NeginDn: I think the most efficient way to write the code is without an if block:
```
var name = "Daniel";
var danielsRole = "mentor";
console.log("Hi, I'm  " + name + ", I'm " + danielsRole);
```
But since the exercise requires an if block, I change it to this:  
```
var name = "Daniel";
var danielsRole = "mentor";
if ((danielsRole === "mentor")) {
  console.log("Hi, I'm Daniel, I'm a mentor.");
} else if (danielsRole === "student") {
  console.log("Hi, I'm Daniel, I'm a student.");
} else {
  console.log("Hi, I'm  " + name + ", I'm " + danielsRole)
}
```

mandatory/1-fix-functions.js
NeginDn: I see, Thank you for letting me know, I've change the code

NeginDn:
Hi Louise,
Thank you so much for all the feedbacks, I have changed my codes according to them. You are the best.

PR #48, opened 2021-01-28
South Africa - Ntshembo Hlongwane

Ntshembo-Hlongwane1:
# Your Details

- Your Name: Ntshembo
- Your City: Cape Town
- Your Slack Name: Ntshembo Hlongwane

# Homework Details

- Module: JS-Core1
- Week: 2

### DIFF
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -9,6 +9,12 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+if (danielsRole === "mentor") {
+  console.log("Hi, I'm Daniel, I'm a mentor.");
+} else {
+  console.log("Hi, I'm Daniel, I'm a student.");
+}
+
 /* 
 EXPECTED RESULT
 ---------------
--- a/exercises/F-logical-operators/exercise.js
+++ b/exercises/F-logical-operators/exercise.js
@@ -11,14 +11,14 @@ var cssLevel = 4;
 
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
-var htmlLevelAbove5;
-var cssLevelAbove5;
+var htmlLevelAbove5 = true;
+var cssLevelAbove5 = false;
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-var cssAndHtmlAbove5;
-var cssOrHtmlAbove5;
+var cssAndHtmlAbove5 = htmlLevel > 5 && cssLevel > 5 ? true : false;
+var cssOrHtmlAbove5 = htmlLevel > 5 || cssLevel > 5 ? true : false;
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/mandatory/3-playing-computer.js
+++ b/mandatory/3-playing-computer.js
@@ -28,7 +28,7 @@ const f2 = function (a, b) {
 
 console.log(x);
 console.log(a);
-console.log(b);
+// console.log(b);
 
 for (let i = 0; i < 5; ++i) {
   a = a + 1;
@@ -39,4 +39,16 @@ for (let i = 0; i < 5; ++i) {
     const e = f1(i, a);
     console.log(e);
   }
+  i;
 }
+
+/**
+  * Answers
+  
+  * 1. Ln31 --> b is not defined hence program throws an error
+  * 3. The value of x, a, d and e is printed
+  * 4. f1 is called once
+  * 5. f2 is called once
+  * 6. value is zero
+  * 7. Value is one
+**/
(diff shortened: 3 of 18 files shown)


exercises/E-conditionals/exercise.js
```diff
@@ -9,6 +9,12 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+if (danielsRole === "mentor") {
```
mulindierick, line 12: `if (danielsRole)`

that works too. remember you've already defined danielsRole
just wanted to share that otherwise, everything else looks pretty

exercises/F-logical-operators/exercise.js
```diff
@@ -11,14 +11,14 @@ var cssLevel = 4;
...
-var cssLevelAbove5;
+var htmlLevelAbove5 = true;
+var cssLevelAbove5 = false;
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-var cssAndHtmlAbove5;
-var cssOrHtmlAbove5;
+var cssAndHtmlAbove5 = htmlLevel > 5 && cssLevel > 5 ? true : false;
```
mulindierick, line 20: `var cssAndHtmlAbove5 = htmlLevelAbove5 && cssLevelAbove5 ` gives you the same result. the ternary operator you've used simplifies it but if you were to use the usual if statements it would lead to longer lines of code.
same applies to the next expression

mandatory/3-playing-computer.js
```diff
@@ -39,4 +39,16 @@ for (let i = 0; i < 5; ++i) {
...
+/**
+  * Answers
+  
+  * 1. Ln31 --> b is not defined hence program throws an error
+  * 3. The value of x, a, d and e is printed
+  * 4. f1 is called once
+  * 5. f2 is called once
+  * 6. value is zero
+  * 7. Value is one
+**/
```
mulindierick, line 54: the loop is running five times,
at 3. the values of f2 and f1 are also printed. remember you've been told to remove the error
at 4 and 5.  f2 will be called thrice for all even numbers and f1 will be called twice for all odd numbers. remember the loop runs five times. f2 is called for only even numbers and f1 is called for only odd numbers 
at 6. so for the first call of f1, the value of a is 8
at 7. but the original value remains 6 

this question was a bit confusing so I spent much time to figure it out. I am not sure if I am entirely correct. think about my suggestions and let me know if you come up with different ideas. you should try to log a few results to prove yourself. 

otherwise, great job

mulindierick, line 54: the loop is running five times,
so, f2 will be called thrice for all even numbers and f1 will be called twice for all odd numbers. remember the loop runs five times. f2 is called for only even numbers and f1 is called for only odd numbers 
at 6. so for the first call of f1, the value of a is 8, remember a is being incremented every time the loop runs but the original value of "a" at the top remains 6 

this question was a bit confusing so I spent much time to figure it out. I am not sure if I am entirely correct. think about my suggestions and let me know if you come up with different ideas. you should try to log a few results to prove yourself. 

otherwise, great job

Ntshembo-Hlongwane1: Oh yea I see what you meaning with regards to **how many times is the f1 called I forgot there's a loop there** thanks for pointing that out...

However with regards to this question below:

> What value does the "a" parameter take in the first "f1" call?

On the first call of f1 they are referring to the **a parameter** and the parameter are they not referring to the parameter on the defined function which in this case is occupied by / getting the value of **i** which on the first run is zero because 
```js
 a = a + 1;
```
That above on the first run will result to 7 failing to hit the **if block** jumping to the **else block** where **i** is zero

**What you think?**

exercises/F-logical-operators/exercise.js
Ntshembo-Hlongwane1: Thanks bro I catch your drift

exercises/E-conditionals/exercise.js
Ntshembo-Hlongwane1: Thanks never thought  of that in that regard....

mandatory/3-playing-computer.js
mulindierick: I see where you are coming from with the 7. There is something I want you to notice though. when the program is run the f2 run first. at this point a is 7. when the loop comes back for the second time, that's when the f1 is run for the first time and a is 8. so it goes like this 

```
step1. run the loop
step2. increment a // now a is 7 because it was initially 6
step3. run f2 ( f2 runs because `i` is 0 and 0 is an even number. at this point the loop goes back without running f1)
step4. go back to running the loop
step5. increment a // now a is 8 
step6. run f1 ( f2 does not run because now `i` is 1 and 1 is an odd number. f1 now runs here for the first time. a will have been incremented twice and a will be 8)

You get it now? Let me know what you think 
```

Ntshembo-Hlongwane1: Nah G I hear you 100% this question was confusing and tricky but yea I understand your point you right

PR #51, opened 2021-01-29
South Africa Class 1 - Jonathan Heugh - JavaScript-Core 1- Week 2

Jontas77:
# Your Details

- Your Name: Jonathan
- Your City: Cape Town
- Your Slack Name: Jonathan

# Homework Details

- Module: JavaScript - Core 1
- Week: 2


KudahShambare:
Good work indeed! 
Just revisit your last question in the mandatory section. I think there is an error there.

Jontas77:
> Good work indeed!
> Just revisit your last question in the mandatory section. I think there is an error there.

Hi Kuda 
Thanks......Are you referring to 3-playing-computer exercise because I did check and there were no errors unless one of my answers is wrong. Can you advise then?

extra/1-radio-stations.js
```diff
@@ -1,97 +1,116 @@
...
+ *
+ * This function should:
+ * - Get the available frequencies from `getAllFrequencies`
+ * - There is a helper function called isRadioFrequency that takes an integer as an argument and returns a boolean.
+ * - Return only the frequencies that are radio stations.
+ */
+// `getStations` goes here
+
+
+function getStations() {
```
Ntshembo-Hlongwane1, line 38: So on this question I doubt this is what they wanted hence I high-key believe this may not be passing all test cases...

### Here's an idea behind this function

1. You have created a function **getAllFrequencies** which returns an array back to you
2. Now what you need from this function called **getStations** you need to get **available frequencies from getAllFrequencies function that you created**
3. Now you can call the **getAllFrequencies** function save the returned array in a variable resulting to you having an array of all frequencies which you returned in that function **getAllFrequencies** 
4. Now they are telling us in the instructions that they made a helper function called **isRadioFrequency** now this function will tell **true** if the frequency is a radio station or **false** if not 
5. Now you have interact with that helper function to help you clean up the array containing all **frequencies** to only remain with frequencies that are for a radio station
6. Finally return all **radio stations**

## Hope this helps 🔥💪

Jontas77: Hi Ntsembo

Thank you for the info and help, I did not finish this but will try to finish this with your advise if time allow it.

KudahShambare: @Jontas . Sorry for that. It's just a difference in answers and I have realised that mine is wrong since many people are going for yours.

ezekiel444 (commented):
check extra-- 1-radio_station: it needs a fix. from Ntshembo comment you are closer to the required result.

Nice work overall.

PR #55, opened 2021-01-29
London Class 7 - Rafael Flores - JavaScript-Core-1-Week 2

rfloresram:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/exercises/D-predicates/exercise.js
+++ b/exercises/D-predicates/exercise.js
@@ -6,10 +6,14 @@
 */
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  return number < 0;
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  return 10 > number > 0;
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -8,6 +8,11 @@
 
 var name = "Daniel";
 var danielsRole = "mentor";
+if ((danielsRole = "mentor")) {
+  console.log("Hi, I'm Daniel, I'm a mentor.");
+} else {
+  console.log("Hi, I'm Daniel, I'm a student.");
+}
 
 /* 
 EXPECTED RESULT
--- a/exercises/F-logical-operators/exercise2.js
+++ b/exercises/F-logical-operators/exercise2.js
@@ -6,6 +6,27 @@
 */
 
 function isNegative() {}
+function isNegative(number1) {
+  let result = number1 < 0;
+  return result;
+}
+
+function isBetween5and10(number1) {
+  let result = number1 <= 10 && number1 > 5;
+  return result;
+}
+
+function isShortName(name1) {
+  let result = name1.length;
+  if (result < 10) return true;
+  else return false;
+}
+
+function startsWithD(name1) {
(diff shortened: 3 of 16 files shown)


exercises/J-array-get-set/exercises2.js
```diff
@@ -6,14 +6,20 @@
   - change the first value in the array to the number 1
 */
 
+/*
 var numbers = [1, 2, 3]; // Don't change this array literal declaration
+numbers[3] = 4;
+*/
+
+var numbers = [1, 2, 3]; // Don't change this array literal declaration
+numbers[numbers.length] = 4;
```
aishak05, line 15: Hi Rafael, great work. Only suggestion I had was you could use the push function here:
numbers.push(4);

exercises/D-predicates/exercise.js
```diff
@@ -6,10 +6,14 @@
...
-function isNegative(number) {}
+function isNegative(number) {
+  return number < 0;
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  return 10 > number > 0;
+}
```
oksygenn, line 16: Seems like your function only checks if number less than 10.
Have you tried testing this function with different number for example 0, -1, 1, 11? 
Try this way, it should work - `10 > number && number > 0`

exercises/F-logical-operators/exercise2.js
```diff
@@ -6,6 +6,27 @@
...
+function isBetween5and10(number1) {
+  let result = number1 <= 10 && number1 > 5;
+  return result;
+}
+
+function isShortName(name1) {
+  let result = name1.length;
+  if (result < 10) return true;
+  else return false;
+}
```
oksygenn, line 23: You could also make this function shorter, as it is not necessary to use if statement if the function is returning a boolean. Because result of comparison is a boolean itself. `return name.length < 10` - returns a boolean.

oksygenn (commented):
Hi Rafael! Well done with the homework. I only have a couple suggestions, but otherwise great job :)

exercises/E-conditionals/exercise.js
```diff
@@ -8,6 +8,11 @@
 
 var name = "Daniel";
 var danielsRole = "mentor";
+if ((danielsRole = "mentor")) {
```
lpalma, line 11: Hey Rafael :wave: 

Try changing the variable `danielsRole` to `"student"` and see what the console prints.
Can we figure out why and how to fix it?

mandatory/1-fix-functions.js
```diff
@@ -10,41 +10,37 @@
 
 */
 
-function mood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function mood(isHappy) {
+  if (isHappy===true) {
```
lpalma, line 14: since `isHappy` is a boolean, we don't need to compare it to `true` again. The original code `if (isHappy)` will also work.

mandatory/2-function-creation.js
```diff
@@ -15,15 +19,24 @@ Complete the function to check if the variable `num` satisfies the following req
...
 
-function validate(num) {}
+/*  function validate(num) {
+  if (typeof num === "number" && num % 2 === 0 && num <= 100){
+    return true; 
+  } else return false;
+}
+*/
+
+const validate = num => (typeof num === "number") && (num % 2 === 0) && (num <= 100) ? true : false;
```
lpalma, line 29: Here we don't need the `? true : false`.
Each boolean expression we have in the code will evaluate to either `true` or `false` and the combination of all these small checks is a boolean expression as well :)

PR #56, opened 2021-01-29
South Africa/Ulrich Ernstzen

ULLIE2298:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -5,7 +5,13 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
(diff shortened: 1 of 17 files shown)


mandatory/2-function-creation.js
```diff
@@ -15,15 +21,24 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
```
Khaya56, line 18: Great work ! @ULLIE2298 , alternatively, to write fewer lines of code you could have had had all your conditions in the if statement.

PR #61, opened 2021-01-29
London class7 ahmad janah

AhmadJanah:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = (studentCount+mentorCount) <= roomMaxCapacity; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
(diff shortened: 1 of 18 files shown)


MSUMBA-art:
nice work buddy, congraturations.

what about in mandatory _3-playing-computer.js_  if you can delete all th b and the last console.log(b) , then run the code and see what will happen

AhmadJanah:
> nice work buddy, congraturations.
> 
> what about in mandatory _3-playing-computer.js_ if you can delete all th b and the last console.log(b) , then run the code and see what will happen

Hi buddy,

Thanks for your comments,

Could you please explain more as I can't find any "console.log(b)" in my code and they asked us to predict without running the code!

Regards,

MSUMBA-art:
The easiest way is to replace all B, with X

And delete the last concole.log(b).

Just delete all B and put X, wherever there's b

exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
...
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = (studentCount+mentorCount) <= roomMaxCapacity; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB; // finish this statement
```
bisioolajide, line 17: Great work Ahmad, there's just a minor point that I noticed, where you used "==", I believe we are advised to always use "===".

PR #68, opened 2021-01-30
London Class 7 - Oksana Kostenko - JavaScript Core 1 - Week 2

oksygenn:
# Your Details

- Your Name: Oksana Kostenko
- Your City: London
- Your Slack Name: Oksana Kostenko

# Homework Details

- Module: JavaScript Core 1
- Week:2

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity === studentCount + mentorCount; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA === personB; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/J-array-get-set/exercise.js
+++ b/exercises/J-array-get-set/exercise.js
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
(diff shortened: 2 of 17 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity === studentCount + mentorCount; // finish this statement
```
ademalagoz, line 13: Great work Oksana. I only suggest this part. Because  yes it can return as true, but it may not be working properly. Please consider below alternatives. Well done.
var enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount;
var enoughSpaceInRoom = roomMaxCapacity >= 25;

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[arr.length - 1]; // complete this statement
```
Hozan94, line 12: Hi Oksana, 
I answered this question the same way as you, but there is another easier method which is `.pop()`

mandatory/2-function-creation.js
```diff
@@ -34,7 +49,32 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(arr) {}
+function formatPercentage(arr) {
+  let newArr = arr.map((num) => {
```
Hozan94, line 53: I like this method, I would not mind if you could explain to me how this works.

Hozan94 (commented):
Hi @oksygenn , I reviewed your work and it is very good, no major suggestions needed from my point of view.
Well done

PR #70, opened 2021-01-30
London Class 7 - Somayeh Karimi - javaScript - core - 1 - Week 2 

DibaKarimi:
# Your Details

- Your Name: Somayeh Karimi
- Your City: London
- Your Slack Name: Somayeh Kr

# Homework Details

- Module: javaScript-core1
- Week: week2

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -8,13 +8,16 @@
 var studentCount = 16;
 var mentorCount = 9;
 var moreStudentsThanMentors; // finish this statement
+moreStudentsThanMentors = studentCount > mentorCount;
 
 var roomMaxCapacity = 25;
 var enoughSpaceInRoom; // finish this statement
+enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount;
 
 var personA = "Daniel";
 var personB = "Irina";
 var sameName; // finish this statement
+sameName = personA === personB;
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-2.js
+++ b/exercises/G-conditionals-2/exercise-2.js
@@ -7,7 +7,14 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if (grade < 50) {
+    return "failed";
+  }
+  else if (grade >= 50) {
+    return "passes"
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-3.js
+++ b/exercises/G-conditionals-2/exercise-3.js
@@ -8,7 +8,17 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80) {
+    return "A";
+  } else if (mark < 80 && mark > 60) {
+    return "B";
+  } else if (mark <= 60 && mark > 50) {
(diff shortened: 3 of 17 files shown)


exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,17 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80) {
+    return "A";
+  } else if (mark < 80 && mark > 60) {
```
NeginDn, line 14: Here in your both else if blocks in line 14 and 16 you wouldn't need to check that the mark is smaller than 80 as it was already checked. Something like: else if (mark > 60) return "B". This is something I just learned from Nigel and is very helpful to know

NeginDn (commented):
Hi Somayeh,
Great work. Very neat codes. I just made few suggestion in case you might find them helpful

exercises/G-conditionals-2/exercise-2.js
```diff
@@ -7,7 +7,14 @@
...
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if (grade < 50) {
+    return "failed";
+  }
+  else if (grade >= 50) {
+    return "passes"
+  }
+}
```
AbdulazizAlhasan, line 17: Good job Somayah. If you only have two options, you do not need to use else if.
in this scenario, the first option was less than 50, that means the other one will definetly by >=50.

you can use
 return "passes"
 without elseif statement, and it will still work

mandatory/1-fix-functions.js
```diff
@@ -21,30 +20,30 @@ function mood() {
...
 function sortArray(letters) {
-  let sortedLetters = letters;
+  let sortedLetters = letters.sort();
 
   return sortedLetters;
 }
 
 function first5(numbers) {
-  let sliced;
+  let sliced=[numbers[0],numbers[1],numbers[2],numbers[3],numbers[4]]
```
AbdulazizAlhasan, line 39: you can use array.slice() function

mandatory/2-function-creation.js
```diff
@@ -5,7 +5,14 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
+  for (let i = 0; i < strArr.length;i++) {
+    strArr[i] = strArr[i].trim();
+    strArr[i] = strArr[i].replace(`/`, ``);
+    strArr[i] = strArr[i].toLowerCase();
```
AbdulazizAlhasan, line 12: good work. you can apply these functions in one go .trim().replace().toLowerCase()

exercises/C-comparison-operators/exercise.js
```diff
@@ -8,13 +8,16 @@
 var studentCount = 16;
 var mentorCount = 9;
 var moreStudentsThanMentors; // finish this statement
+moreStudentsThanMentors = studentCount > mentorCount;
 
 var roomMaxCapacity = 25;
 var enoughSpaceInRoom; // finish this statement
+enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount;
```
jmodiale, line 15: Great Job so far Somayeh. Just to make the code shorter as much as possible - you can use 
enoughSpaceInRoom = roomMaxCapacity>=25.. rather than repeating since it's more or less saying the same thing

rfloresram:
Good syntax and handling in order of your code, congratulations keep it up!

PR #74, opened 2021-01-30
London Class 7 - Benyam Worku - HTML/CSS - Week 2

BenyamWorku:
# Your Details

- Your Name: Benyam Worku
- Your City: London
- Your Slack Name: Benyam Worku

# Homework Details

- Module: Javascript Core 1 
- Week: Week 2

### DIFF
--- a/exercises/G-conditionals-2/exercise-1.js
+++ b/exercises/G-conditionals-2/exercise-1.js
@@ -6,7 +6,9 @@
   - if number is more or equal to zero, return the word "positive"
 */
 
-function negativeOrPositive(number) {}
+function negativeOrPositive(number) {
+  return number >= 0 ? "positive" : "negative";
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/I-array-properties/exercise.js
+++ b/exercises/I-array-properties/exercise.js
@@ -6,7 +6,12 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  if (arr.length === 0) {
+    return true;
+  } else {
+    return false;
+  }
+  // short way would be return !arr.length;
 }
 
 /* 
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -32,18 +32,12 @@ function getAllFrequencies() {
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
-function isRadioFrequency(frequencies) {
-  // if (getAllFrequencies().includes(frequencies)) {
-  //   return true;
-  // }
-  return getAllFrequencies().includes(frequencies);
-}
-function getStations() {
-  
-  
-  }
 
-  
+function getStations() {
+  //
+  return getAllFrequencies().filter(isRadioStation);
+  // another way would be getAllFrequencies().filter((station) => isRadioStation(station));
+}
 
 /* ======= TESTS - DO NOT MODIFY ======= */
 
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -14,6 +14,13 @@
  */
 
 // `getAllFrequencies` goes here
+function getAllFrequencies() {
+  let freqArray = [];
+  for (let i = 87; i <= 108; i++) {
+    freqArray.push(i);
+  }
+  return freqArray;
+}
 
 /**
  * Next, let's write a function that gives us only the frequencies that are radio stations.
@@ -25,6 +32,18 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
+function isRadioFrequency(frequencies) {
+  // if (getAllFrequencies().includes(frequencies)) {
+  //   return true;
+  // }
+  return getAllFrequencies().includes(frequencies);
+}
+function getStations() {
+  
+  
+  }
+
+  
 
 /* ======= TESTS - DO NOT MODIFY ======= */
 
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -18,10 +18,8 @@ function mood(isHappy) {
   }
 }
 
-function greaterThan10(num) {
-  let isBigEnough = 10;
-
-  if (num > isBigEnough) {
+function greaterThan10(isBigEnough) {
+  if (isBigEnough > 10) {
     return "num is greater than 10";
   } else {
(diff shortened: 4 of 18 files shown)


exercises/G-conditionals-2/exercise-1.js
```diff
@@ -6,7 +6,9 @@
   - if number is more or equal to zero, return the word "positive"
 */
 
-function negativeOrPositive(number) {}
+function negativeOrPositive(number) {
+  return number >= 0 ? "positive" : "negative";
+}
```
oksygenn, line 11: Nice use of ternary operator! 👍  You could also write this function in one line by declaring it as an arrow function, you wouldn't even need to use `return` and curly brackets  :)

exercises/I-array-properties/exercise.js
```diff
@@ -6,7 +6,12 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  if (arr.length === 0) {
+    return true;
+  } else {
+    return false;
+  }
```
oksygenn, line 13: You could also make this function shorter, as it is not necessary to use if statement if the function is returning a boolean. Because result of comparison is a boolean itself. `return arr.length === 0` - returns a boolean.

oksygenn (commented):
Hi Benyam! Well done with the homework 👍 I learned/remembered a couple of things from your code. I left some minor suggestions below

mandatory/1-fix-functions.js
```diff
@@ -21,30 +19,30 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = 10;
```
amellie, line 22: Your solution is not wrong, but the variable name is called `isBigEnough` so you should do the test if the number is bigger than 10 here :)

mandatory/2-function-creation.js
```diff
@@ -5,7 +5,12 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
+  for (let i = 0; i < strArr.length; i++) {
+    strArr[i] = strArr[i].trim().replace("/", "").toLowerCase();
```
amellie, line 10: Well done on doing the trim, replacement, and coversion to lower case, all in one line!

mandatory/2-function-creation.js
```diff
@@ -15,17 +20,31 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  let conditions = typeof num === "number" && num % 2 === 0 && num <= 100;
```
amellie, line 24: Well done! Another way is to simplify the condition by removing the check if it's a number. ``num % 2 === 0`` will return false if ``num`` is not a number OR not even

mandatory/2-function-creation.js
```diff
@@ -15,17 +20,31 @@ Complete the function to check if the variable `num` satisfies the following req
...
+  // The below code is the longer solution
```
amellie, line 25: I wouldn't suggest a longer solution since the condition is pretty straightforward. We would even do ``return num % 2 === 0 && num <= 100;`` without saving it into a variable. It is desirable if the condition is not complex :)

mandatory/3-playing-computer.js
```diff
@@ -26,17 +26,17 @@ const f2 = function (a, b) {
...
+console.log(a); // this a is a global variable
+// console.log(b); This line throws the error as b is not defined
 
 for (let i = 0; i < 5; ++i) {
-  a = a + 1;
+  a = a + 1; // a is incremented 5 times as long as the loop's index is less that 5
   if (i % 2 === 0) {
-    const d = f2(i, x);
-    console.log(d);
+    const d = f2(i, x); // f2 is called at i=0,2,4
```
amellie, line 36: perfect!

extra/1-radio-stations.js
```diff
@@ -26,6 +33,12 @@
  */
 // `getStations` goes here
 
+function getStations() {
+  //
+  return getAllFrequencies().filter(isRadioStation);
```
amellie, line 38: very well done!

mandatory/2-function-creation.js
```diff
@@ -15,17 +20,31 @@ Complete the function to check if the variable `num` satisfies the following req
...
 
 /* 
 Write a function that returns a copy of the given array arr, but with the element at the given index, index removed.
 The function must NOT change the original array, arr.
 */
 
 function remove(arr, index) {
-  return; // complete this statement
+  let newArray = [];
+  for (let i = 0; i < arr.length; i++) {
```
amellie, line 41: Since our last discussion, I found a way to do this in a simpler manner using `splice` and `slice`. We can discuss this in our meeting :)

exercises/I-array-properties/exercise.js
amellie: I agree, the shorter way is better as it is readable :) good effort!

amellie:
I can see that you have mastered the essential parts of JS :) Keep up the good work!

amellie (changes requested):

mandatory/2-function-creation.js
BenyamWorku: I am not sure what you mean. Can it get any simpler than this ?

mandatory/2-function-creation.js
BenyamWorku: Ahh! got it now. No need to check the data type of the number.

mandatory/2-function-creation.js
BenyamWorku: Sure we can discuss it. I have done it that way too.

BenyamWorku:
> I can see that you have mastered the essential parts of JS :) Keep up the good work!

Thanks Amalia for reviewing my work

mandatory/2-function-creation.js
amellie: In your solution above, you don't actually need `slice`. Try to check what `splice` does

BenyamWorku: Thanks done it using splice . Splice modifies the array in place hence the need for a deep copy of the array. And one thing I learnt about it is just like pop() and shift() splice () returns the item that has been removed at index. Live and learn!

PR #75, opened 2021-01-31
London Class 7 - Soraiya Muhammad - JavaScript Core 1 - Week 2 

soraiya:
# Your Details

- Your Name: Soraiya Muhammad
- Your City: London
- Your Slack Name: soraiya

# Homework Details

- Module: JS Core 1
- Week: 2

### DIFF
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -5,7 +5,16 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+
+let tidyUpString = strArr => strArr.map(elem => elem.trim().replace('/', '').toLowerCase());
+// function tidyUpString(strArr) {
+//    let tidyArr = [];
+//    for(let i = 0; i < strArr.length; i++) {
+//       strArr[i].trim().replace('/', '').toLowerCase();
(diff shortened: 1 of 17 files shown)


mandatory/3-playing-computer.js
```diff
@@ -28,7 +28,6 @@ const f2 = function (a, b) {
 
 console.log(x);
 console.log(a);
-console.log(b);
```
Domenico-cyf, line 31: Well done Soraiya, just looking as the common way to solve the issue. However you could assign B a value of zero and you would have had similar outcome. Truth is, we are not 100% sure of if there was any B or if it has been missed.
Overall great code.

mandatory/2-function-creation.js
```diff
@@ -34,7 +47,17 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(arr) {}
+function formatPercentage(arr) {
+  let newArr = [];
+  for (let i = 0; i < arr.length; i++) {
+  let = numChange = Number(arr[i].toFixed(2));
```
DouglasVDM, line 53: Your code looks great and easy to read.
I think you have too many = signs when declaring the numChange variable.
Was it intentional?
If yes, what was the reason?

PR #85, opened 2021-05-16
Jest-ify tests

illicitonion:
There are also several modifications to tests here:
* mandatory/1-fix-functions.js no longer has `sortArray` or `first5`
  exercises - we moved array methods to week 3, so the tests should move
  there too.
* mandatory/2-function-creation.js has some array-based tests changed to
  instead test the same function creation but for single values. We
  moved teaching `map` to week 3, I'll re-add the array versions of
  these tests to week 3.
* extra/1-radio-stations.js changes how we populate the expected
  stations list, because the tests were previously failing if duplicate
  stations were generated by the random generator.

### DIFF
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -40,11 +40,11 @@ function getAvailableStations() {
     while (getAvailableStations.stations.length < stationCount) {
       let randomFrequency = Math.floor(Math.random() * (108 - 87 + 1) + 87);
       if (!getAvailableStations.stations.includes(randomFrequency)) {
-          getAvailableStations.stations.push(randomFrequency);
+        getAvailableStations.stations.push(randomFrequency);
       }
-    };
-    getAvailableStations.stations.sort(function(frequencyA, frequencyB) {
-        return frequencyA - frequencyB
+    }
+    getAvailableStations.stations.sort(function (frequencyA, frequencyB) {
+      return frequencyA - frequencyB;
     });
   }
 
@@ -57,28 +57,8 @@ function isRadioStation(frequency) {
 
 test("getAllFrequencies() returns all frequencies between 87 and 108", () => {
   expect(getAllFrequencies()).toEqual([
-    87,
-    88,
-    89,
-    90,
-    91,
-    92,
-    93,
-    94,
-    95,
-    96,
-    97,
-    98,
-    99,
-    100,
-    101,
-    102,
-    103,
-    104,
-    105,
-    106,
-    107,
-    108,
+    87, 88, 89, 90, 91, 92, 93, 94, 95, 96, 97, 98, 99, 100, 101, 102, 103, 104,
+    105, 106, 107, 108,
   ]);
 });
 
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -28,19 +28,23 @@
 
 /* ======= TESTS - DO NOT MODIFY ======= */
 
+const {expect, test} = require("@jest/globals");
+
 function getAvailableStations() {
   // Using `stations` as a property as defining it as a global variable wouldn't
   // always make it initialized before the function is called
   if (!getAvailableStations.stations) {
     const stationCount = 4;
-    getAvailableStations.stations = new Array(stationCount)
-      .fill(undefined)
-      .map(function () {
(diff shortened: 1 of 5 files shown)


mandatory/2-function-creation.js
```diff
@@ -17,93 +17,66 @@ Tip: use logical operators
...
+
+test("validate function accepts other valid even number", () => {
+    expect(validate(18)).toEqual(true);
+});
+
+test("validate function accepts exactly 100", () => {
+    expect(validate(100)).toEqual(true);
+});
+
+test("validate function works rejects odd number", () => {
```
40thieves, line 59: Typo in the title here

```suggestion
test("validate function rejects odd number", () => {
```

mandatory/2-function-creation.js
```diff
@@ -17,93 +17,66 @@ Tip: use logical operators
...
+    expect(validate(17)).toEqual(false);
+});
+
+test("validate function works rejects string", () => {
```
40thieves, line 63: Same here

```suggestion
test("validate function rejects string", () => {
```

mandatory/2-function-creation.js
```diff
@@ -17,93 +17,66 @@ Tip: use logical operators
...
+    expect(validate("Ten")).toEqual(false);
+});
+
+test("validate function works rejects stringified number", () => {
```
40thieves, line 67:
```suggestion
test("validate function rejects stringified number", () => {
```

40thieves (commented):
A few minor issues, but overall looks good! Thanks for taking this on @illicitonion!

extra/1-radio-stations.js
```diff
@@ -28,19 +28,23 @@
 
 /* ======= TESTS - DO NOT MODIFY ======= */
 
+const {expect, test} = require("@jest/globals");
+
```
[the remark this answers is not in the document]
illicitonion: Done!

mandatory/1-fix-functions.js
```diff
@@ -51,80 +39,45 @@ function get3rdIndex(arr) {
...
-);
-test(
-  "get3rdIndex function works - case 2",
-  get3rdIndex([11, 37, 62, 18, 19, 3, 30]),
-  18
-);
+const {expect, test} = require("@jest/globals");
+
+test("mood function works for true", () => {
+    expect(mood(true)).toEqual("I am happy");
```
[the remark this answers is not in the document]
illicitonion: Ugh, half of our existing stuff was 2-space and half was 4-space. I've just run prettier over them all.

mandatory/1-fix-functions.js
```diff
@@ -51,80 +39,45 @@ function get3rdIndex(arr) {
...
+});
+
+test("mood function works for false", () => {
+    expect(mood(false)).toEqual("I am not happy");
+});
+
+test("greaterThanTen function works for value greater than 10", () => {
+    expect(greaterThan10(11)).toEqual("num is greater than 10");
```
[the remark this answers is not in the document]
illicitonion: I added an extra one - tbh though I think the "fix basic syntax" and "fix functions" tests really are better judged by human eyes than tests - there's very specific _exact source code_ we're expecting the trainees to write here, rather than a more vague specification we're giving them for the function...

mandatory/2-function-creation.js
```diff
@@ -1,11 +1,11 @@
 /*
 Write a function that:
-- takes an array of strings as input
-- removes any spaces in the beginning or end of the strings
-- removes any forward slashes (/) in the strings
+- takes a string as input
```
ChrisOwen101, line 3: This probably makes more sense as a task however as this in JS1-2 it makes sense that they should be working with Arrays in this task. 

Is it too much trouble to change it back to using arrays?

ChrisOwen101 (commented):
Great work here - just one comment

mandatory/2-function-creation.js
illicitonion: Done - I'll update the next week's homework to also say:
```
NOTE: This exercise is the same as one you did last week - try to do it again using things you learnt this week.
Think about what is better about this solution than your one last week, and what is worse.
```
because I think it's worth repeating when they know `map`.

PR #87, opened 2021-06-28
Northwest-class4-AnzaAzam-Javascript Core1-Week2-Mandatory 

Anza-Azam:

# Your Details

- Your Name: Anza Azam
- Your City: North West
- Your Slack Name: Anza-Azam

# Homework Details

- Module: Javascript Core1
- Week: 1

# Notes

- What did you find easy?
functions
- What did you find hard?
string methods

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,10 +10,10 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(mood) {
+  let isHappy=mood;
 
-  if (isHappy) {
+  if (isHappy===true) {
     return "I am happy";
   } else {
     return "I am not happy";
@@ -21,9 +21,9 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough=num;
 
-  if (isBigEnough) {
+  if (isBigEnough>10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
@@ -32,7 +32,7 @@ function greaterThan10(num) {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element=arr[index];
 
   return element;
 }
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -15,7 +15,7 @@ function validate(num) {
   else return false;
 
   }
-console.log(validate(18));
+
 
 
 
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -6,7 +6,18 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  let number = num;
+  if (typeof number === 'number' && number <= 100 && (number%2 === 0)) {
+    return true;
+  }
+  
+  else return false;
+
+  }
+console.log(validate(18));
+
+
 
 /*
 Write a function that:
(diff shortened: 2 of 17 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,20 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(mood) {
+  let isHappy=mood;
 
-  if (isHappy) {
+  if (isHappy===true) {
```
KarenPudner, line 16: You could remove`===true` and the code will still work.

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,18 @@ Complete the function to check if the variable `num` satisfies the following req
...
-function validate(num) {}
+function validate(num) {
+  let number = num;
+  if (typeof number === 'number' && number <= 100 && (number%2 === 0)) {
+    return true;
+  }
+  
+  else return false;
+
+  }
```
KarenPudner, line 17: This is fine. You could remove the if/else and do this in one line.
`return(some expression)`
will return true or false depending on if the statement is true or false.

mandatory/2-function-creation.js
```diff
@@ -16,7 +27,20 @@ Write a function that:
...
+  if (num > 100) { num = 100; }
+  if (Number.isInteger(num)) {
+    num = num;
+  }
+  else { num = num.toFixed(2); }
+      num = `${num}%`;
+  if (num.includes("0") && num.indexOf("0") === 4) {
+    num = num.slice(0, 4);
+    num = `${num}%`;
+
```
KarenPudner, line 40: This will only remove a single trailing zero. So it will change 18.10 to 18.1. But it won't change 18.00 to 18, for example. This is quite tricky, and there are a few different ways of doing it. Do some googling about removing trailing zeros in javascript, and decide which method is best.

mandatory/2-function-creation.js
```diff
@@ -25,7 +49,22 @@ Write a function that:
...
+  for (let i = 0; i < array.length; i++) {
+    array[i] = array[i].trim();
+    array[i] = array[i].toLowerCase();
+
+    if (array[i].includes('/')) {
+
+      index = array[i].indexOf('/');
+      array[i] = array[i].slice(index + 1, array[i].length);
+
+    }
```
KarenPudner, line 63: There is a simpler way of doing this, with the String replace() method.

mandatory/3-playing-computer.js
```diff
@@ -1,42 +1,67 @@
...
+ *   5. How many times is "f2" called?
+ *   3 times
+ *   6. What value does the "a" parameter take in the first "f1" call?
+ * 
+ *   "a" parameter value will be 1 because it is i in f1 first call
+ * 
+ *   7. What is the value of the "a" outer variable when "f1" is called for the first time?
+ * 
+ *   The value of a is 8
+ *
```
KarenPudner, line 39: Yes, these are all correct

mandatory/2-function-creation.js
Anza-Azam: thanks, Karen, just used replace now as replace('/',''). replace is so simple.

mandatory/2-function-creation.js
Anza-Azam: I've now changed it to parseFloat thanks, otherwise, I wouldn't have known about it. Thanks, much appreciated.
	num = parseFloat(num.toFixed(2));

mandatory/1-fix-functions.js
Anza-Azam: Changes made.thanks

mandatory/2-function-creation.js
Anza-Azam: Changed it as return( expression)

PR #88, opened 2021-06-28
Nw-class4-myaskylar-Js-core1-wk2

myaskylar:

# Your Details

- Your Name: Mya Skylar
- Your City: NW
- Your Slack Name: Mya Skylar

# Homework Details

- Module: Js core 1
- Week: Wk 2

# Notes

- What did you find easy?
  Array, Comparison Operator, Logical operator
- What did you find hard?
 To understand the questions.
Needs more exercises.
- Any other notes?

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = studentCount + mentorCount <= roomMaxCapacity; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -14,7 +14,7 @@ var enoughSpaceInRoom = studentCount + mentorCount <= roomMaxCapacity; // finish
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName = personA == personB; // finish this statement
+var sameName = personA === personB; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,7 +5,36 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(number) {
+  if (number < 0) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+function isBetween5and10(num) {
+  if (num >= 5 && num <= 10){
+    return true;
+  } else {
+    return false;
+  }
+}
+
+function isShortName(name) {
+  if (name === "Daniel"){
+    return true;
+  } else {
+    return false;
+  }
+  
+}
+
+function startsWithD(nameStartWithD) {
+   
+    return nameStartWithD.startsWith("D");
+  
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
@@ -24,5 +53,5 @@ console.log("Does Daniel start with 'D'?", startsWithD("Daniel"));
   Is 5 a negative number? false
   Is 10 in the range 5-10? true
   Is Daniel a short name? true
-  Does Daniel start with 'D'?
+  Does Daniel start with 'D'? true
 */
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -22,7 +22,7 @@ function isBetween5and10(num) {
 }
 
 function isShortName(name) {
-  if (name === "Daniel"){
+  if (name.length <= 6){
     return true;
   } else {
     return false;
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -6,28 +6,16 @@
 */
 
 function isNegative(number) {
-  if (number < 0) {
-    return true;
-  } else {
-    return false;
-  }
+ return number < 0;
 }
 
 function isBetween5and10(num) {
-  if (num >= 5 && num <= 10){
-    return true;
-  } else {
-    return false;
-  }
+  return num >= 5 && num <= 10;
 }
 
 function isShortName(name) {
-  if (name.length <= 6){
-    return true;
-  } else {
-    return false;
-  }
   
+  return name.length <= 6;
 }
 
 function startsWithD(nameStartWithD) {
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -10,10 +10,10 @@ var name = "Daniel";
 var danielsRole = "mentor";
 
 function createGreetingForDaniel(role) {
-  if (role === danielsRole ) {
+  if (role === "mentor" ) {
     return `Hi, I'm ${name}, I'm a ${danielsRole}`;
   } else {
-    return `Hi, I'm ${name}, I'm a student`;
+    return `Hi, I'm ${name}, I'm a ${danielsRole}`;
   }
 }
 
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -9,6 +9,16 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+function createGreetingForDaniel(role) {
+  if (role === danielsRole ) {
+    return `Hi, I'm ${name}, I'm a ${danielsRole}`;
+  } else {
+    return `Hi, I'm ${name}, I'm a student`;
+  }
+}
+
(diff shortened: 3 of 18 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
...
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = studentCount + mentorCount <= roomMaxCapacity; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB; // finish this statement
```
KarenPudner, line 17: This is correct. You could also use === 
Do you know the difference between == and === in javascript? If not then try and find out.

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,36 @@
...
+    return false;
+  }
+}
+
+function isShortName(name) {
+  if (name === "Daniel"){
+    return true;
+  } else {
+    return false;
+  }
```
KarenPudner, line 29: I think the name of the function indicates that you should be checking name length.

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,36 @@
...
+  
+}
+
+function startsWithD(nameStartWithD) {
+   
+    return nameStartWithD.startsWith("D");
+  
+}
```
KarenPudner, line 37: In this function you have used just one line. In the other methods you have done 2 separate checks. Both  are correct. Which do you prefer? Or does it depend? There's no right answer, but it's good to think about.

exercises/E-conditionals/exercise.js
```diff
@@ -9,6 +9,16 @@
...
 var danielsRole = "mentor";
 
+function createGreetingForDaniel(role) {
+  if (role === danielsRole ) {
+    return `Hi, I'm ${name}, I'm a ${danielsRole}`;
+  } else {
+    return `Hi, I'm ${name}, I'm a student`;
+  }
+}
+
```
KarenPudner, line 19: I think you should hardcode "mentor" in your function rather than use danielsRole.

exercises/F-predicates/exercise.js
```diff
@@ -6,10 +6,14 @@
...
-function isNegative(number) {}
+function isNegative(number) {
+ return number < 0 === true;
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  return number > 0 && number < 10 === true;
+}
```
KarenPudner, line 16: Do you need === true? Will the code still work without it?

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,17 @@
...
+  if (mark >= 80){
+    return "A";
+  } else if (mark < 80 && mark > 60){
+    return "B";
+  } else if (mark < 60 && mark > 50){
+    return "C";
+  } else {
+    return "F"
+  }
+}
```
KarenPudner, line 21: This works, but you could simplify the code. If the mark is greater or equal to 80, then you code will exit at line 13. So you don't need to do the check
mark < 80
on line 14. If the mark must be less than 80, or your code would already have exited.
Similarly, you can simplify line 16

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,13 @@
...
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+  if (sentence.startsWith("code")){
+    return true;
+  } else {
+    return false;
+  }
+}
```
KarenPudner, line 17: The question asks you to check if a sentence contains the word "code" not if it starts with "code" You need a function that will also return true for a sentence like "I am learning to code"

exercises/C-comparison-operators/exercise.js
myaskylar: Thanks Karen. Yes  == is used for comparison between two variables irrespective of the datatype of variable. === is used for comparison between two variables but this will check strict type, which means it will check datatype and compare two values.

exercises/D-logical-operators/exercise2.js
myaskylar: Thanks Karen. Predicate function takes one value as input and return it true or false. I've changed the code to simplified it into one line. I think this will be good practice. Thanks for the feedback. Its helps me to improvise my code.

mandatory/1-fix-functions.js
```diff
@@ -10,31 +10,27 @@
 
 */
 
-function mood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function mood(isHappy) {
+  if (isHappy === true) {
```
KarenPudner, line 14: You can remove`===true`, 
`if (isHappy)` will work on its own.

mandatory/1-fix-functions.js
```diff
@@ -10,31 +10,27 @@
...
 }
 
 function get3rdIndex(arr) {
-  let index = 3;
-  let element;
+  let index = arr.slice(3, 4);
+  let element = index;
 
-  return element;
+  return element[0];
```
KarenPudner, line 33: I think it would be easier to use `return arr[x]`
where x is the correct index number.

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,13 @@ Complete the function to check if the variable `num` satisfies the following req
...
 */
 
-function validate(num) {}
+function validate(num) {
+  if (num % 2 == 0 && num <= 100 && typeof(num) === "number"){
+    return true;
+  } else {
+    return false;
+  }
+}
```
KarenPudner, line 15: This works fine. You could also do it without using if/else. 
`return(some expression)`
will return true or false depending on whether the expression is true or false

mandatory/2-function-creation.js
```diff
@@ -25,7 +40,14 @@ Write a function that:
...
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+  for(i = 0; i < arrayOfStrings.length; i++){
+    arrayOfStrings[i] = arrayOfStrings[i].replace("/", "");
+    arrayOfStrings[i] = arrayOfStrings[i].trim();
+    arrayOfStrings[i] = arrayOfStrings[i].toLowerCase();
+  }
+ return arrayOfStrings;
+}
```
KarenPudner, line 50: This is fine. You could also chain the 3 functions into 1 line of code, but this is personal preference - either way is correct.

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,51 @@
...
+      6
+      4
+      9
+      6
+      13
+      8
   4. How many times is "f1" called?
+     3times
   5. How many times is "f2" called?
+     2times
```
KarenPudner, line 24: Check your answers to 4 and 5

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,51 @@
...
   6. What value does the "a" parameter take in the first "f1" call?
+     8
```
KarenPudner, line 26: This is a confusing question, but it means what is the value of a in the line 
`const f1 = function (a, b)`
when it is called for the first time. I can see from your commented out code below that you do know what this is.

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,51 @@
...
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
+     6
```
KarenPudner, line 28: Again, it's a confusing question. It means, what is the value of a in the line
` const e = f1(i, a);`
when it is called for the first time. And. again, I can see from your comments on that line that you know what this is.

mandatory/1-fix-functions.js
myaskylar: Thanks Karen. I understand this part now. I've change it.

mandatory/2-function-creation.js
myaskylar: Thanks Karen I've tried the method and its work! cool!

mandatory/2-function-creation.js
myaskylar: I've changed it to one line of code.

exercises/C-comparison-operators/exercise.js
KarenPudner: Yes, that's exactly right.

extra/1-radio-stations.js
```diff
@@ -14,7 +14,11 @@
  */
 
 // `getAllFrequencies` goes here
-
+function getAllFrequencies() {
+  const frequencies = [87,88,89,90,91,92,93,94,95,96,97,98,99,100,101,102,103,104,105,106,107,108]
+  return frequencies ;
```
KarenPudner, line 19: You should do this using a for loop. It's simple enough to hardcode the array here as it's quite short, but imagine if you had an array with hundreds or thousands of values -  a for loop would be your only practical option.

extra/1-radio-stations.js
```diff
@@ -25,6 +29,26 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
+function getStations() {
+
+ let stations = getAllFrequencies();
+ 
+ let availableStations = [];
+ for (let i = 0; i < stations.length; i++) {
+   if (isRadioStation(stations[i])){
```
KarenPudner, line 38: Maybe assign frequency[i] to a variable called something like currentStation, to make the code more understandable.

PR #91, opened 2021-06-28
NorthWest - Class 4 - Anza Azam -Javascript Module Core 1 -Week 2 -Extra challenge

Anza-Azam:

# Your Details

- Your Name: Anza Azam
- Your City: North west
- Your Slack Name:Anza-Azam

# Homework Details

- Module: JS Core1
- Week: week 2

# Notes

- What did you find easy?
pseudo code function modelling
- What did you find hard?
finding builtin methods

- Any other notes?

### DIFF
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -17,9 +17,12 @@
 
 function getAllFrequencies() {
 	let frequencies = [];
-	for (let i = 87; i < 109; i++) {
+	let frequencyStart = 87;
+	let frequencyEnd = 109;
+
+	for (let i = frequencyStart; i < frequencyEnd; i++) {
 		frequencies.push(i);
-		}
+	}
 	return frequencies;
 }
 
@@ -37,14 +40,12 @@ function getAllFrequencies() {
 // `getStations` goes here
 
 function getStations() {
-
-  
-  let station = [];
-
-	let frequency= getAllFrequencies();
+	let station = [];
+	let frequency = getAllFrequencies();
 	for (let i = 0; i < frequency.length; i++) {
-		if (isRadioStation(frequency[i])) {
-			station.push(frequency[i]);
+		validFrequency = frequency[i];
+		if (isRadioStation(validFrequency)) {
+			station.push(validFrequency);
 		}
 	}
 	return station;
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -2,6 +2,8 @@
  * Finding a radio station, and a good one, can be hard manually.
  * Let's use some code to help us build a program that helps us scan
  * the radio waves for some good music.
+ *
+ * @format
  */
 
 /**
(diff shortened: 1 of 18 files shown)


erin-switchstitch (commented):
I really liked your use of both the lower case 'd' and the uppercase 'D' for exercise 2 !

erin-switchstitch (commented):
Exercises G-conditionals - exercise 3 :
Not particularly important but you may be able to reduce the amount of code in your 'else if' statements. If you set them out in the right order (which you have) you would not need to say 
"  else if (mark < 80 && mark > 60) "       only      "else if (mark >= 60) "


  if (mark >= 80) {
    return "A";
  } else if (mark >= 60) {
    return "B"
  } else if (mark >= 50) {
    return "C";
  } else {
    return "F";
  }}

erin-switchstitch (commented):
I've gone through and checked all of your exercises (B to J) and as far as I can see everything there is correct - I've left a couple of comments about some very slight tweaks you may want to make. Well done, really good job :) !!!!

extra/1-radio-stations.js
```diff
@@ -13,6 +15,14 @@
  * - Should return this array to use in other functions
  */
 
+function getAllFrequencies() {
+	let frequencies = [];
+	for (let i = 87; i < 109; i++) {
+		frequencies[i - 87] = i;
+	}
+	return frequencies;
+}
```
konikalily, line 24: Hi Anza, just a thought : could you consider using .push() here instead of frequencies[i - 87] = i ? Otherwise your code is very good. Thank you.

konikalily (changes requested):
Hi Anza, just a thought : could you consider using .push() here instead of frequencies[i - 87] = i ? Otherwise your code is very good. Thank you.

extra/1-radio-stations.js
```diff
@@ -13,6 +15,14 @@
  * - Should return this array to use in other functions
  */
 
+function getAllFrequencies() {
+	let frequencies = [];
+	for (let i = 87; i < 109; i++) {
```
KarenPudner, line 20: Maybe assign 87 and 109 to variables called something like startFrequency and endFrequency to make the code more understandable

extra/1-radio-stations.js
```diff
@@ -26,42 +36,58 @@
  */
 // `getStations` goes here
 
+function getStations() {
+
+  
+  let station = [];
+
+	let frequency= getAllFrequencies();
```
KarenPudner, line 44: This shouldn't be indented - it should be the same indentation as `let station = [];`

extra/1-radio-stations.js
```diff
@@ -26,42 +36,58 @@
...
+	for (let i = 0; i < frequency.length; i++) {
+		if (isRadioStation(frequency[i])) {
```
KarenPudner, line 46: Maybe assign `frequency[i]` to a variable called something like currentFrequency, to make the code more understandable.

PR #96, opened 2021-06-29
North West 4 - Gulnihal Naldoken- JS Core 1 - Week 2

gulnihaln:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/F-predicates/exercise.js
+++ b/exercises/F-predicates/exercise.js
@@ -6,11 +6,15 @@
 */
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  return number < 0;
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
-
+function isBetweenZeroAnd10(number) {
+  return number >= 0 && number <= 10;
+}
+  
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -14,6 +14,16 @@
  */
 
 // `getAllFrequencies` goes here
+function getAllFrequencies(){
+  // const arrAllFrequencies =[87, 88, 89, 90, 91, 92, 93, 94, 95, 
+  //                       96, 97, 98, 99, 100, 101, 102, 103, 104,
+  //                       105, 106, 108, 108];
+  const arrAllFrequencies = [];
+  for (let i = 87; i< 109; i++){
+    arrAllFrequencies.push(i);
+  }
+  return arrAllFrequencies;
+}
 
 /**
  * Next, let's write a function that gives us only the frequencies that are radio stations.
@@ -25,7 +35,16 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
-
+function getStations() {
+  let frequencies = getAllFrequencies();
+  let availableStations = [];
+  for (i = 0; i < frequencies.length; i++){
+    if (isRadioStation(frequencies[i])){
+      availableStations.push(frequencies[i]);
+    }
+  }
+  return availableStations;
+}
 /*
  * ======= TESTS - DO NOT MODIFY =======
  * Note: You are not expected to understand everything below this comment!
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,10 +10,9 @@
 
 */
 
-function mood(moods) {
-  let isHappy = true;
-
-  if (moods === isHappy) {
+function mood(isHappy) {
+  
+  if (isHappy) {
     return "I am happy";
   }else {
     return "I am not happy"
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,31 +10,28 @@
 
 */
 
-function mood() {
+function mood(happy) {
   let isHappy = true;
 
-  if (isHappy) {
+  if (happy === isHappy) {
     return "I am happy";
-  } else {
-    return "I am not happy";
+  }else {
+    return "I am not happy"
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = false;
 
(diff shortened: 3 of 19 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,31 +10,28 @@
 
 */
 
-function mood() {
+function mood(happy) {
   let isHappy = true;
 
-  if (isHappy) {
+  if (happy === isHappy) {
```
konikalily: Hi Gulnihaln, your overall code is good. However I think you can write only " if (isHappy)"  if you have (isHappy) as the parameter of the mood function instead of having a separated variable. Ex:
function mood(isHappy) {
  if (isHappy) {
    return "I am happy";
  } else {
 "I am not happy"
}
}

konikalily (changes requested):
Hi Gulnihaln, your overall code is good. However I think you can write only " if (isHappy)"  if you have (isHappy) as the parameter of the mood function instead of having a separated variable. Ex:
function mood(isHappy) {
  if (isHappy) {
    return "I am happy";
  } else {
 "I am not happy"
}
}

mandatory/1-fix-functions.js
gulnihaln: Thank you Konika, I don't know the problem but this is what I tried and failed. I've tried it again now, it failed again.

gulnihaln: Thanks Konika, I tried it before but I've found my mistake now, when I wrote it in this way.

exercises/F-predicates/exercise.js
```diff
@@ -6,11 +6,15 @@
 */
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
```
Abed-Aghbar, line 9: Hi Gulnihaln, 
Good work.
Just here I think you need to look again to your good .
function isNegative(number){
  if (number === 5){
    return false;
  } else {
      return true;
  }
}

exercises/F-predicates/exercise.js
```diff
@@ -6,11 +6,15 @@
...
+  return number < 0;
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
-
+function isBetweenZeroAnd10(number) {
+  return number >= 0 && number <= 10;
```
Abed-Aghbar, line 15: Here as well
function isBetweenZeroAnd10(number){
  if (number === 5){
    return true;
}   else {
      return false;
}
}

Abed-Aghbar (commented):
Good work

exercises/F-predicates/exercise.js
gulnihaln: Hi Abed, Thanks for your review. I think I understood what you meant. If I'm right, when I set the function as in your example, it will only work for 5. 5 is  just a test number, it could be anything. Now, it works for any negative number.

exercises/F-predicates/exercise.js
gulnihaln: Hi, thanks. Yes. same here. If I create function in this way it will be running correctly for 5, but I want it be true for all numbers "BetweenZeroAnd10".

extra/1-radio-stations.js
```diff
@@ -14,6 +14,16 @@
  */
 
 // `getAllFrequencies` goes here
+function getAllFrequencies(){
+  // const arrAllFrequencies =[87, 88, 89, 90, 91, 92, 93, 94, 95, 
+  //                       96, 97, 98, 99, 100, 101, 102, 103, 104,
+  //                       105, 106, 108, 108];
+  const arrAllFrequencies = [];
+  for (let i = 87; i< 109; i++){
```
zeinaji, line 22: To avoid magic numbers here, it's worth having a variable for `startFrequency` and `endFrequency`

extra/1-radio-stations.js
```diff
@@ -25,7 +35,16 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
-
+function getStations() {
+  let frequencies = getAllFrequencies();
```
zeinaji, line 39: This could be a `const` as it won't change from now on

zeinaji (commented):
This is great as always, the logic is sound throughout and the extension task is well done 😊

extra/1-radio-stations.js
gulnihaln: That one was really challenging for me. I am going to try to understand what you've meant.

mandatory/1-fix-functions.js
```diff
@@ -10,31 +10,27 @@
...
-  } else {
-    return "I am not happy";
+  }else {
+    return "I am not happy"
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = false;
```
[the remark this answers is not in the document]
gulnihaln: Ohh yes, I corrected the other ones that I've solved in this way. I missed that one. will do it.

extra/1-radio-stations.js
zeinaji: The logic is done perfectly so you definitely don't need to worry about that 🙂, I just mean instead of having the numbers 87 and 109 in the function without explaining what they're for, it may be better to assign them to variables, so have 
```
const startFrequency = 87;
const endFrequency = 109;
```

mandatory/2-function-creation.js
```diff
@@ -16,7 +18,13 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+  if (typeof num === "number" && num <= 100){
+  return `${Math.trunc(num*100)/100}%`;
```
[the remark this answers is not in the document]
gulnihaln: I've tried with it but couldn't log the answer.

extra/1-radio-stations.js
gulnihaln: I got it now, thank you. I was studying on freecodecamp, I made smt like this in an exercise just now.

PR #108, opened 2021-07-02
NW1- Erin Dyson - Javascript_1 - Week_2

erin-switchstitch:
Exercises and Mandatory completed. No work done on extra yet

# Your Details

- Your Name: Erin Dyson
- Your City: Manchester
- Your Slack Name: erin-switchstitch

# Homework Details

- Module: JSS 1 
- Week: Week 2 

# Notes

- What did you find easy?
All of the exercises and 1 and 2 of mandatory
- What did you find hard?
3rd mandatory challenge
n/a
- Any other notes?

-----
[View rendered exercises/A-expressions/README.md](https://github.com/erin-switchstitch/JavaScript-Core-1-Coursework-Week2/blob/main/exercises/A-expressions/README.md)
[View rendered exercises/F-predicates/README.md](https://github.com/erin-switchstitch/JavaScript-Core-1-Coursework-Week2/blob/main/exercises/F-predicates/README.md)

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -5,16 +5,16 @@
   The variables should have values that match the expected results.
 */
 
-var studentCount = 16;
+var studentCount = 15;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity == studentCount + mentorCount
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,7 +5,47 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(number) {
+
+  if (number < 0){
+    return true
+  } else {
+    return false
+  }
+
+}
+
+function isBetween5and10(number) {
+
+  if ((number >= 5) && (number <= 10)) {
+    return true
+  } else {
+    return false
+  }
+
+}
+
+function isShortName(name) {
+
+  if ( name.length <= 6 ) {
+    return true
+  } else {
+    return false
+  }
+
+}
+
+function startsWithD(name) {
+
+  for ( i = 0 ; i < name.length ; i++){
+      if (name[0] == "D") {
+        return true;
+      } else {
+        return false; 
+      }
+  }
+} 
+
 
 /* 
   DO NOT EDIT BELOW THIS LINE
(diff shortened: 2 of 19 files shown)


Sharm-Dev:
Really nice work Erin, it was nice to see your visual explanation of what was happening to the code every time it was ran in the loop in the Playing Computer exercise.

exercises/C-comparison-operators/exercise.js
```diff
@@ -5,16 +5,16 @@
...
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity == studentCount + mentorCount
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB; // finish this statement
```
peedeerich, line 17: There are two operators for checking equality in JavaScript - "==" and "==="

"==" is loose equality - here JavaScript will convert types when checking
"===" is strict equality - here JavaScript will not convert the types, so is true only if they are the same type.

For example - see what happens if you try `1 == "1"` versus `1 === "1"`

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,47 @@
...
+
+  for ( i = 0 ; i < name.length ; i++){
+      if (name[0] == "D") {
+        return true;
+      } else {
+        return false; 
+      }
+  }
+} 
+
```
peedeerich, line 48: This will work just fine but you don't need the loop; you can just check `name[0] == "D"`. The `return` statement here will mean that the loop exits after the first step - once a function hits a `return` it will stop executing

You can simplify it even more as well! `name[0] == "D"` will itself be either `true` or `false` so you can return it directly.

exercises/F-predicates/exercise.js
```diff
@@ -6,10 +6,22 @@
 */
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  if (number < 0) {
+    return true;
+  } else {
+    return false; 
+  }
```
peedeerich, line 14: as before - you could return `number < 0` here

exercises/I-array-properties/exercise.js
```diff
@@ -6,8 +6,11 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
-}
+  if (arr[0] == undefined) {
+    return true; // complete this statement
+} else {
+  return false
+}}
```
peedeerich, line 13: This could fail sometimes - for example it would class the array `[undefined, 'Erin', 'Paul']` as empty. You could check the length as an alternative

mandatory/2-function-creation.js
```diff
@@ -25,7 +48,22 @@ Write a function that:
...
+    var toEdit = arrayOfStrings[i];
+    var toTrim = toEdit.trim();
+    var toReplace = toTrim.replace('/','');
+    var toLower = toReplace.toLowerCase();
+    freshArray.push(toLower);
+  }
+  return freshArray;
+}
+
+
```
peedeerich, line 66: very nice - you may have some more ideas how you could do this following the most recent class (hint: `map` could help)

peedeerich (commented):
Excellent work :) I've left a few comments but nothing major - reach out if there is anything which you'd like to discuss about these exercises

PR #125, opened 2021-07-11
JS1 week 2 NW4 Khadar M Dagal

khmdagal:


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,22 +7,24 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount + mentorCount; // finish this statement
```
zeinaji, line 10: `moreStudentsThanMentors` should equal either true of false, with having `studentCount + mentorCount` here you'll end up with a number instead

exercises/C-comparison-operators/exercise.js
```diff
@@ -7,22 +7,24 @@
...
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = moreStudentsThanMentors; // finish this statement
```
zeinaji, line 13: This also should be a boolean, so either true or false

exercises/D-logical-operators/exercise.js
```diff
@@ -11,29 +11,56 @@ var cssLevel = 4;
 
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
-var htmlLevelAbove5;
```
zeinaji, line 14: Here you could have `htmlLevelAbove5 = htmlLevel > 5` and this would behave the same as the `htmlLevelAbove5` function. This applies to all the other variables in here, so `cssLevelAbove5`, `cssAndHtmlAbove5` and `cssOrHtmlAbove5`. So for example you could also have `cssAndHtmlAbove5 = htmlLevel > 5 && cssLevel > 5`

exercises/D-logical-operators/exercise2.js
```diff
@@ -4,8 +4,39 @@
...
+
+function isBetween5and10(num) {
+    if (num > 5 && num <= 10) {
+        return true;
+    }
+    return false;
+}
+
+function isShortName(name) {
+    if (name === "Daniel") {
```
zeinaji, line 25: Here you want the name to be short, it doesn't necessarily have to be Daniel, so you need to find out the length of the name

exercises/D-logical-operators/exercise.js
```diff
@@ -11,29 +11,56 @@ var cssLevel = 4;
...
+        return true;
+    }
+    return false;
+}
 
-/* 
-  DO NOT EDIT BELOW THIS LINE
-  --------------------------- */
 
-console.log("Is Html knowledge above 5?", htmlLevelAbove5);
```
zeinaji, line 27: If you do it like the comment above, it should work with these expected results 🙂

exercises/E-conditionals/exercise.js
```diff
@@ -9,8 +9,12 @@
 var name = "Daniel";
 var danielsRole = "mentor";
 
+if (danielsRole === "mentor") {
+    console.log("Hi, I'm " + name + ", I'm " + danielsRole + ".");
+}
```
zeinaji, line 14: You need an `else if` block here in case `danielsRole === "student"`

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,12 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+    if (sentence.includes("code")) {
```
zeinaji, line 12: The logic here is correct, but to make this more concise you could have 
```
     return sentence.includes("code")
}
```
This should work the same

exercises/H-array-literals/exercise.js
```diff
@@ -4,8 +4,8 @@
   Declare some variables assigned to arrays of values
 */
 
-var numbers = []; // add numbers from 1 to 10 into this array
-var mentors; // Create an array with the names of the mentors: Daniel, Irina and Rares
+var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]; // add numbers from 1 to 10 into this array
```
zeinaji, line 7: Here you want to create a function that adds these numbers to any array. So something like 
```
function addNumbersToArray(array){

}
```
Using a `for` loop here could help

exercises/H-array-literals/exercise.js
```diff
@@ -4,8 +4,8 @@
...
+var mentors = ["Daniel", "Irina", "Rares"]; // Create an array with the names of the mentors: Daniel, Irina and Rares
```
zeinaji, line 8: Same thing here 🙂

zeinaji (commented):
Nice work Khadar. Just thought I'd have a look at the exercises section for now and leave the mandatory one to go over on Tuesday, hope that's okay. Let me know if you have any questions 🙂

mandatory/2-function-creation.js
```diff
@@ -6,86 +6,130 @@ Complete the function to check if the variable `num` satisfies the following req
...
+    if (Number.isInteger(num)) {
+        return num + "%";
+    }
+
+
+    var rounding = Math.round(num);
+    var n = num.toFixed(2);
+    var numberPlusePrecentage = n + '%';
+    var twodicemalpalce = numberPlusePrecentage;
+    return numberPlusePrecentage;
```
zeinaji, line 39: The logic here is good, the only thing I'd say is make sure you use either `let` or `const` instead of `var` because it's outdated. Also, as an alternative, you could use `Math.round(num * 100) / 100`, this way you don't need an `if` block for when the number is an integer, cause this works on it too https://stackoverflow.com/questions/11832914/how-to-round-to-at-most-2-decimal-places-if-necessary

zeinaji: Also you could get rid of the `rounding` variable because it's not used anywhere 🙂

PR #127, opened 2021-07-15
Scotland Class 5 - Ivan Zhukov - JS Module 1 week 2

zhkvivan:

# Your Details

- Your Name: Ivan Zhukov
- Your City: Glasgow
- Your Slack Name: Ivan Zhukov

# Homework Details

- Module: JS core 1
- Week: 2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = 16 > 9; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = 25 >= studentCount + mentorCount; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA === personB; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-2.js
+++ b/exercises/G-conditionals-2/exercise-2.js
@@ -7,7 +7,13 @@
 
 */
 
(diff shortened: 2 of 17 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = 16 > 9; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = 25 >= studentCount + mentorCount; // finish this statement
```
sriram-siv, line 13: This is all good logic, but you could improve this by referring to the saved variables.

moreStudentsThanMentors = studentCount > mentorCount
enoughSpaceInRoom = roomMaxCapacity >=  studentCount + mentorCount

This way, if you ever need to update a number like maxCapacity then you only have to change it in one place.

exercises/G-conditionals-2/exercise-2.js
```diff
@@ -7,7 +7,13 @@
...
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+	if (grade < 50) {
+		return 'failed'
+	} else if (grade >= 50) {
+		return 'passed'
+	}
+}
```
sriram-siv, line 16: If a number is not less than 50 it is always more than or equal to 50 instead. So you can remove the second condition here as it will always be true.

if (grade < 50) {
    return "failed"
} else {
    return "passed
}

exercises/C-comparison-operators/exercise.js
zhkvivan: Yeah that's a good point.

exercises/G-conditionals-2/exercise-2.js
zhkvivan: Thank you, this does make sense.

PR #152, opened 2021-08-31
West Midlands Class 3 - Selahaddin Alan - JS - Week 2

justsealan:

# Your Details

- Your Name:Selahaddin Alan
- Your City: Nottingham
- Your Slack Name:Selahaddin Alan

# Homework Details

- Module:JavaScript 
- Week:2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


exercises/J-array-get-set/exercises2.js
```diff
@@ -7,7 +7,7 @@
 */
 
 var numbers = [1, 2, 3]; // Don't change this array literal declaration
-
+numbers[3] = 4;
 /*
```
tenzyns, line 11: There is a red cross on the top of this commit. I think may be you need to assign numbers[0] = 1; as asked in the question.

mandatory/1-fix-functions.js
```diff
@@ -10,8 +10,8 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(mood){
```
tenzyns, line 13: I think there should be a Boolean expression that evaluates to true or false. And I think mood can stay as function name, and isHappy as parameter. Then an if statement may work as follow-->> (isHappy === true) return 'I am happy'.

mandatory/2-function-creation.js
```diff
@@ -25,7 +39,13 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+  //return arrayOfStrings.map(string => string.trim().replace("/ ", '').toLowerCase());
+  for (let i = 0; i < arrayOfStrings.length; i++) {
+    arrayOfStrings[i] = arrayOfStrings[i].trim().replace("/", "").toLowerCase();
```
tenzyns, line 45: This was done nicely. I got stuck here and needed help in applying these array modification.

tenzyns:
I have completed the review. Work well done, Selahaddin.

exercises/J-array-get-set/exercises2.js
justsealan: Isn't the first value in the array already the number 1. Do I need to change the number 1 with number 1? Why? I don’t get it 😃

mandatory/1-fix-functions.js
justsealan: Can you write the code you mean? I don’t get it.

exercises/J-array-get-set/exercises2.js
tenzyns: yes i agree. the question was bit silly. But that's what the question asks us to do. may be that will remove the red cross on your file.

mandatory/1-fix-functions.js
tenzyns: If you check your "if statement", how does 'isHappy' become a boolean expression without any boolean operators. Also you have both function name and parameter name the same ---mood

PR #170, opened 2021-12-19
London Class 8 - Matilda Ako - JS1 - Week 2

MatildaAko:

# Your Details

- Your Name: Matilda Ako
- Your City: London
- Your Slack Name: Matilda Ako - trainee

# Homework Details

- Module: JS1
- Week: 2

# Notes

- What did you find easy? Most of the work was very simple, the only exception was the extra task which was a little tricky 

- What did you find hard? Working with the helper function, I realised I was using the wrong array for the parameter

- Any other notes?

### DIFF
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,7 +5,21 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(num) {
+    return num < 0 ? true : false;
+}
+
+function isBetween5and10(num) {
+    return num >= 5 && num <= 10 ? true : false;
+}
+
+function isShortName(name) {
+    return name.length <= 6 ? true : false;
+}
+
+function startsWithD(name) {
+    return name[0] === "D" ? true : false;
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-4.js
+++ b/exercises/G-conditionals-2/exercise-4.js
@@ -8,7 +8,9 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+    return sentence.includes("code") ? true : false;
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -6,7 +6,9 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+    return Number.isInteger(num) && num % 2 === 0 && num <= 100 ? true : false;
+}
 
 /*
 Write a function that:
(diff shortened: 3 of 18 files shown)


MatildaAko:
Hi @haack I've completed this week's work

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,21 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(num) {
+    return num < 0 ? true : false;
```
IwonaKrysp, line 9: It's a nice shortcut, is that mean the same as 
return true and then return false?

mandatory/2-function-creation.js
```diff
@@ -25,67 +34,74 @@ Write a function that:
...
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+    let newArray = [];
+    for (var i = 0; i < arrayOfStrings.length; i++) {
+        let arr = arrayOfStrings[i].trim().toLowerCase().replace("/", "");
+        newArray.push(arr);
+    }
+    return newArray;
+}
```
IwonaKrysp, line 45: very nice and neat function

mandatory/3-playing-computer.js
```diff
@@ -6,37 +6,36 @@
...
-}
+    a = a + 1;
+    if (i % 2 === 0) {
+        const d = f2(i, x);
+        console.log(d);
+    } else {
+        const e = f1(i, a);
+        console.log(e);
+    }
+}
```
IwonaKrysp, line 41: Your code is so nice and neat, very pleasant to look at.

exercises/D-logical-operators/exercise2.js
MatildaAko: Yes, this is a ternary operator, it's a short way of writing an if statement. You can find out about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)

mandatory/3-playing-computer.js
MatildaAko: Thank you 😊

exercises/D-logical-operators/exercise2.js
haack: You've done this below but just as a reminder you can actually just return the boolean part of this directly like this without the ternary:
```suggestion
    return num < 0;
```

The reason for this is because `num < 0` actually is already equal to `true` or `false`. With the ternary it is like saying: `if (num < 0 === true) { return true } ....`

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,9 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+    return sentence.includes("code") ? true : false;
```
haack, line 12: Same comment as before about not needing the ternary operator

haack (commented):
Awesome 👍

PR #171, opened 2021-12-19
London8_Mandeep_Sangha_JS1_wk2

mandeepsangha:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount ; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA === personB; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -5,16 +5,16 @@
   The variables should have values that match the expected results.
 */
 
-var studentCount = 16;
-var mentorCount = 9;
-var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
+let studentCount = 16;
+let mentorCount = 9;
+let moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
-var roomMaxCapacity = 25;
-var enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount ; // finish this statement
+let roomMaxCapacity = 25;
+let enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount ; // finish this statement
 
-var personA = "Daniel";
-var personB = "Irina";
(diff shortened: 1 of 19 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -5,16 +5,16 @@
   The variables should have values that match the expected results.
 */
 
-var studentCount = 16;
-var mentorCount = 9;
-var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
+let studentCount = 16;
```
teniolao, line 8: Good code layout Mandeep, it's great you were able to pass all the test.

teniolao (commented):
Good code layout Mandeep, it's great you were able to pass all the test.

extra/1-radio-stations.js
```diff
@@ -15,6 +15,15 @@
 
 // `getAllFrequencies` goes here
 
+function getAllFrequencies(){
+  let array = [];
+    for (i = 87; i < 109; i++) {
```
TimeaReich, line 20: Hi! I am reviewing your code, and it looks like you know a lot more than me, so I don't think I can add anything, but i sure learned a few things going through your code so thank you! Nice use of for loop here.

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,18 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
```
MatenRehimi, line 11: Function works correctly but some of the conditions in the if statements are not needed. Try to have a look at which ones might not be need to simplify logic here.

exercises/I-array-properties/exercise.js
```diff
@@ -5,8 +5,13 @@
 
 */
 
-function isEmpty(arr) {
-  return; // complete this statement
+function isEmpty(arr) { // complete this statement
+  if (arr.length === 0){
```
MatenRehimi, line 9: Could replace the if-else with a predicate similar to the function below
function isNegative(number) {
  return number < 0;
}

mandatory/2-function-creation.js
```diff
@@ -25,7 +39,20 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(array) {
```
MatenRehimi, line 42: For this function, it's possible to call multiple methods when returning. So you could have el.trim().toLowerCase().replace(/\//g, '')

extra/1-radio-stations.js
```diff
@@ -25,6 +34,11 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
+function getStations(){
+let array2 = getAllFrequencies().filter(x => isRadioStation(x));
```
MatenRehimi, line 38: Good use of filter here to produce clean/readable code.

exercises/E-conditionals/exercise.js
```diff
@@ -9,6 +9,14 @@
 let name = "Daniel";
 let danielsRole = "mentor";
 
+if (danielsRole = "mentor"){
```
MatenRehimi, line 12: Comparisons should be ===, in this instance you've used = as assignment and given the variable danielsRole the value "mentor again. 

Be careful as danielsRole = "mentor" will still passes the tests as assignments with strings results to true if the length of the string is greater than 1. Here's an example to show this:
console.log(Boolean((names = "h")));   //true
console.log(Boolean((names = "")));    //false

PR #189, opened 2022-01-01
London 8 - Irina Shilova - JavaScript Core 1 Coursework - Week 2 

IrinShilova:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy? Exercises were easy

- What did you find hard? The 3rd task in Mandatory 

- Any other notes? I make a lot of mistakes in syntax


IrinShilova:
@moneyinthesky  Just letting you know that it's done

mandatory/2-function-creation.js
```diff
@@ -25,7 +36,20 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+  for (var i = 0; i < arrayOfStrings.length; i++) {
```
moneyinthesky, line 40: Would be better to use let here instead of var :)

extra/1-radio-stations.js
```diff
@@ -15,6 +15,14 @@
 
 // `getAllFrequencies` goes here
 
+function getAllFrequencies () {
+  let tmpFreq = [];
+  for (i=87; i <= 108; i++) {
```
moneyinthesky, line 20: It looks like you're using i here without declaring it. It's a good idea to use let here, this will limit the scope of the variable to the for loop. We will discuss more about variable scopes in the JS3 module.

extra/1-radio-stations.js
```diff
@@ -26,6 +34,25 @@
  */
 // `getStations` goes here
 
+function getStations () {
+  let allStations = getAllFrequencies(); 
+  let availableStations = [];
+
+  for (i=0; i<allStations.length; i++)
```
moneyinthesky, line 41: Same comment as above about using i without declaring it.

moneyinthesky:
I made a few comments here - overall, this looks very good. I noticed there were 2 files called 1-radio-stations.js, so I only reviewed the one in the extras folder.

extra/1-radio-stations.js
IrinShilova: Yes, I see. I just forgot about it, I didn't do it for purpose.

PR #190, opened 2022-01-03
London Class 8 - Craig D'Silva - Javascript Core 1 - Week 2

craig-dsilva:

# Your Details

- Your Name: Craig D'Silva
- Your City: London
- Your Slack Name: Craig - Trainee

# Homework Details

- Module:JS1
- Week: 2

# Notes

- What did you find easy?
  Fix functions and playing computer were easy

- What did you find hard?
  Function creation and radio stations

  How to use the helper function (isRadioStation) in 1-radio-stations.js

- Any other notes?

### DIFF
--- a/exercises/D-logical-operators/exercise.js
+++ b/exercises/D-logical-operators/exercise.js
@@ -11,14 +11,14 @@ let cssLevel = 4;
 
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
-let htmlLevelAbove5;
-let cssLevelAbove5;
+let htmlLevelAbove5 = htmlLevel >= 5;
+let cssLevelAbove5 = cssLevel >= 5;
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-let cssAndHtmlAbove5;
-let cssOrHtmlAbove5;
+let cssAndHtmlAbove5 = cssLevel >= 5 && htmlLevel >= 5;
+let cssOrHtmlAbove5 = cssLevel >= 5 || htmlLevel >= 5;
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-2.js
+++ b/exercises/G-conditionals-2/exercise-2.js
@@ -7,7 +7,13 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if (grade >= 50) {
+    return "passed";
+  } else if (grade <= 50) {
+    return "failed";
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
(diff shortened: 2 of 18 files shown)


exercises/D-logical-operators/exercise.js
```diff
@@ -11,14 +11,14 @@ let cssLevel = 4;
...
+let htmlLevelAbove5 = htmlLevel >= 5;
+let cssLevelAbove5 = cssLevel >= 5;
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-let cssAndHtmlAbove5;
-let cssOrHtmlAbove5;
+let cssAndHtmlAbove5 = cssLevel >= 5 && htmlLevel >= 5;
+let cssOrHtmlAbove5 = cssLevel >= 5 || htmlLevel >= 5;
```
DomVinyard, line 21: I guess if the variable name is 'above' it should just be `>` rather than `>=` ?

exercises/G-conditionals-2/exercise-2.js
```diff
@@ -7,7 +7,13 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if (grade >= 50) {
+    return "passed";
+  } else if (grade <= 50) {
+    return "failed";
+  }
```
DomVinyard, line 15: is says `if the grade is less than 50` - so we can just say `grade < 50` instead of `<=`

even better we can just use a plain else statement because it the result is inclusive (everything that doesn't match the first argument will match the second:

```
 if (grade >= 50) {
     return "passed";
   } else {
     return "failed";
   }
```

even better we don't need the else at all, because the `return` on line 12 will terminate the function if it matches, so we can reduce this all down to:

```
 if (grade >= 50)  return "passed"
 return "failed"
```

or if you're being extra fancy go for a ternary and fit it all on one line:
```
return (grade >= 50) ? "passed" : "failed"
```

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,17 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80) {
+    return "A";
+  } else if (mark < 80 && mark > 60) {
```
DomVinyard, line 14: we don't need `mark < 80` here, everything above will be returned on the line above so this will never be greater than 80.

exercises/I-array-properties/exercise.js
```diff
@@ -6,7 +6,11 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  if (arr.length > 0) {
+    return false;
+  } else if (arr.length <= 0) {
```
DomVinyard, line 11: an array length cannot be less than zero, here again we can just use an else

```
 if (arr.length > 0) return false
else return true
```

mandatory/1-fix-functions.js
```diff
@@ -10,19 +10,18 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(mood) {
```
DomVinyard, line 13: this is not good naming, `mood` is the name of both the function and the argument, this is confusing

PR #197, opened 2022-01-05
London Class 8 - Ezgi Gunes - JS core 1 - Week 2 

Ezgi3406:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,7 +5,15 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(){
+  if(isNegative=-10){
+    return true;
+  }
+    return false;
+}
+
+   
+
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-2.js
+++ b/exercises/G-conditionals-2/exercise-2.js
@@ -7,7 +7,13 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if(grade>=50){
+    return "passed";
+  }else{
+    return "failed";
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/J-array-get-set/exercise.js
+++ b/exercises/J-array-get-set/exercise.js
@@ -5,11 +5,13 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr.shift(); // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  // return arr.pop()
+  return arr[arr.length-1];
+  ; // complete this statement
 }
 
 /* 
--- a/exercises/J-array-get-set/exercises2.js
+++ b/exercises/J-array-get-set/exercises2.js
@@ -7,6 +7,7 @@
 */
 
 let numbers = [1, 2, 3]; // Don't change this array literal declaration
+let end=numbers.push(4);
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -15,6 +15,14 @@
 
 // `getAllFrequencies` goes here
 
+const getAllFrequencies = () =>{
+  let frequency =[]
+    for (let i = 87; i < 109; i++) {
+      frequency.push(i)
+    }
+    return frequency
+  }
+console.log(getAllFrequencies())
 /**
  * Next, let's write a function that gives us only the frequencies that are radio stations.
  * Call this function `getStations`.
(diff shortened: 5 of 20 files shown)


exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,15 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(){
+  if(isNegative=-10){
+    return true;
+  }
+    return false;
+}
```
mahri09, line 13: Heeey, Ezgi first of all I appreciate your work well done! As we need to review someone code, this time I select your one, and let me add some more points. Here is a task function that return numbers negatives

exercises/G-conditionals-2/exercise-2.js
```diff
@@ -7,7 +7,13 @@
...
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if(grade>=50){
+    return "passed";
+  }else{
+    return "failed";
+  }
+}
```
mahri09, line 16: yeah for just for learning code we can do if-else statement like this you did great job but in the coding we use else if any of our options are not true , then we give else to leave some messages like this: 
if (grade < 50) {
    return " failed";
  } else if (grade >= 50) {
    return " passed";
  } else {
    return " not a valid grade";
  }

exercises/J-array-get-set/exercises2.js
```diff
@@ -7,6 +7,7 @@
 */
 
 let numbers = [1, 2, 3]; // Don't change this array literal declaration
+let end=numbers.push(4);
```
mahri09, line 11: here we don't assign it to a variable we just push it on our existing array by name numbers like this :
numbers.push(4);

mandatory/1-fix-functions.js
```diff
@@ -9,10 +9,8 @@
   NOTE:Only make edits inside the function
 
 */
-
-function mood() {
-  let isHappy = true;
-
+let isHappy = true;
```
mahri09, line 12: no need to add isHappy variable here we take isHappy from parameter when we call this function we gave true or false as an argument, under the tests see they already given arguments to this function

mandatory/1-fix-functions.js
```diff
@@ -21,9 +19,9 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough=true
```
mahri09, line 22: also here no need assign true or false it is come from arguments

mandatory/1-fix-functions.js
```diff
@@ -32,7 +30,7 @@ function greaterThan10(num) {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element= arr[index]
```
mahri09, line 33: Sometimes writing too much code can be unnecessary and takes up too much space. so you can just do
 let element= arr[3];
return element;

mahri09 (commented):
thank you Ezgi your code is clear and well done I enjoyed reading your code!

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,13 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr.shift(); // complete this statement
```
jstadnik, line 8: While this returns the first element, "shift" also removes that element from the array. Since an array is a javascript object, it is passed by reference (that means the function can change what you pass to it), meaning if you called this function, it would remove the first element from the function, which may be quite an unexpected side-effect.

```javascript
function first(arr) {
    return arr[0]
}
```

May be better here.

exercises/J-array-get-set/exercises2.js
jstadnik: According to the exercise it should be
```
numbers[3] = 4
```
Although
```
numbers.push(4)
```
is honestly probably better.

You don't need to assign it to a variable in any case, though.

extra/1-radio-stations.js
```diff
@@ -24,13 +32,25 @@
...
-// `getStations` goes here
 
+   // const isRadioStation =(num)=>(num >= 87 && num <= 108)?true:false
+    //   console.log(isRadioStation(9))
+
+// `getStations` goes here 
+// 
+
+const getStations = () => {
+  return getAllFrequencies().filter((element)=>isRadioStation(element))
```
jstadnik, line 43: Well done!

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,15 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  if(typeof(num)!=='string' && num%2==0 && num<=100){
```
jstadnik, line 10: `typeof(num) == 'number'` may be better to account for other possibilities (e.g `false` would meet all the conditions here), but otherwise :+1:

jstadnik (commented):
Great work!

PR #203, opened 2022-01-06
London Class_8 - Katarzyna Ruksza - JS Core1 - Week_2

KatarzynaRuksza:
# Your Details

- Your Name: Katarzyna Ruksza
- Your City: London
- Your Slack Name: Kate Ruksza (trainee)
 
# Homework Details

- Module: JS Core 1
- Week: 2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 let studentCount = 16;
 let mentorCount = 9;
-let moreStudentsThanMentors; // finish this statement
+let moreStudentsThanMentors = true; // finish this statement
 
 let roomMaxCapacity = 25;
-let enoughSpaceInRoom; // finish this statement
+let enoughSpaceInRoom = true; // finish this statement
 
 let personA = "Daniel";
 let personB = "Irina";
-let sameName; // finish this statement
+let sameName = false; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/D-logical-operators/exercise.js
+++ b/exercises/D-logical-operators/exercise.js
@@ -11,14 +11,14 @@ let cssLevel = 4;
 
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
-let htmlLevelAbove5;
-let cssLevelAbove5;
+let htmlLevelAbove5 = true;
+let cssLevelAbove5 = false;
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-let cssAndHtmlAbove5;
-let cssOrHtmlAbove5;
+let cssAndHtmlAbove5 = false;
+let cssOrHtmlAbove5 = true;
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,7 +5,33 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(num) {
+  if (num < 0) {
+    return true;
+  }
+  return false;
+}
+
+function isBetween5and10(num) {
+  if (num >= 5 && num <= 10) {
+    return true;
+  }
+  return false;
+}
+
+function isShortName(name) {
+  if (name.length <= 6) {
+    return true;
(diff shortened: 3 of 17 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 let studentCount = 16;
 let mentorCount = 9;
-let moreStudentsThanMentors; // finish this statement
+let moreStudentsThanMentors = true; // finish this statement
```
moneyinthesky, line 10: I think the idea here was to use the comparison operators (like >, <, >=, <=, ===) to complete the statement. You could reference the variables that are already being initialised here (studentCount and mentorCount) - so even if the values of those variables change, moreStudentsThanMentors will still have the correct value.

exercises/D-logical-operators/exercise.js
```diff
@@ -11,14 +11,14 @@ let cssLevel = 4;
 
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
-let htmlLevelAbove5;
-let cssLevelAbove5;
+let htmlLevelAbove5 = true;
```
moneyinthesky, line 14: Same comment as above.

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,33 @@
   Update the code so that you get the expected result.
 */
```
moneyinthesky, line 7: These functions look good to me :) Well done!

exercises/E-conditionals/exercise.js
```diff
@@ -9,6 +9,12 @@
 let name = "Daniel";
 let danielsRole = "mentor";
 
+if (danielsRole === "mentor") {
+  console.log("Hi, I'm Daniel, I'm a mentor.");
```
moneyinthesky, line 13: The logic here is good. But it might be a good idea to use the name variable here instead of repeating "Daniel" in the code. Imagine, if we wanted to change the name, we could just update a single variable instead of having to update everywhere in the code that mentions "Daniel"

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,19 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+if (mark >= 80) {
+    return "A";
+  }
+  if (mark < 80 && mark > 60) {
```
moneyinthesky, line 15: This is technically correct. However, just so you know, you could also write this using `else if`.

mandatory/2-function-creation.js
```diff
@@ -16,7 +23,16 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+  if (typeof num === "number") {
```
moneyinthesky, line 27: Here, we already know that the argument to this function will be a number as it's in the description above, so we don't really need to check it again.

mandatory/2-function-creation.js
```diff
@@ -16,7 +23,16 @@ Write a function that:
...
+    if (num > 100) {
```
moneyinthesky, line 28: The rest of the logic in the function looks good to me!

mandatory/2-function-creation.js
```diff
@@ -25,7 +41,11 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+  return arrayOfStrings.map((array) =>
```
moneyinthesky, line 45: We'll cover the Array map() method more in Week 4. The logic here looks correct to me. Something to mention here is that the parameter of the function being passed to the map method is called `array`, but it's not really an array, it's actually a string in this case.

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,33 @@
...
 
-function isNegative() {}
+function isNegative(num) {
+  if (num < 0) {
```
moneyinthesky, line 9: Whenever you have this kind of logic: `if(expression) return true else return false` - you could simplify it by returning the expression directly: `return expression` - as the expression will then be evaluated to true or false anyway.

PR #208, opened 2022-01-07
London Class 8 - Anderson Ferreira - Javascript - week2

andersonuk:
- Your Name: Anderson
- Your City: London
- Your Slack Name: andersonuk

- What did you find easy? 
Some Exercises folder 

- What did you find hard? 
 Mandatory

- Any other notes?


mandatory/2-function-creation.js
```diff
@@ -16,76 +17,85 @@ Write a function that:
...
 - removes any spaces in the beginning or end of each string
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
-
+function tidyUpStrings(arrayOfStrings) {
+    let updatedArray = []
+    for (let idx = 0; idx < arrayOfStrings.length; idx++) {
+        updatedArray.push(arrayOfStrings[idx].trim().replace('/', '').toLowerCase())
```
ewintram, line 36: great use of string functions and a `for` loop

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,43 @@
...
+     b has not been defined
   2. Remove the line that throws the error.
   3. What is printed to the console?
+    2, 6, 4, 9, 6, 13, 8
   4. How many times is "f1" called?
+    Two times
   5. How many times is "f2" called?
+    Three times
   6. What value does the "a" parameter take in the first "f1" call?
+    8
```
ewintram, line 19: this exercise is very unclear because it reuses the same variable names. If you had a look at the solutions you will see this answer is 1 because the `a` the question is referring to is the parameter in the function definition for `f1`, i.e. this `a`
```
const f1 = function (a, b) {
```

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,43 @@
...
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
+    6
```
ewintram, line 21: `a` is originally assigned a value of 6.
On the first iteration of the for loop:
- 1 is added to `a` so its new value is 7
- `f1` is not called because the `if` condition on line 42 is false
On the second iteration of the for loop:
- 1 is added to `a` so its new value is 8
- `f1` is called because the `if` condition on line 42 is true, therefore `f1` is called with the value of `a` set to 8

ewintram (commented):
Excellent job Anderson!

exercises/C-comparison-operators/exercise.js
```diff
@@ -7,29 +7,29 @@
...
+let moreStudentsThanMentors = true; // finish this statement
 
 let roomMaxCapacity = 25;
-let enoughSpaceInRoom; // finish this statement
+let enoughSpaceInRoom = true; // finish this statement
 
 let personA = "Daniel";
 let personB = "Irina";
-let sameName; // finish this statement
+let sameName = false; // finish this statement
```
Hatice06, line 17: Hi Anderson. Great job! You can also use comparison operators (=== same value) like this;

let sameName = personA === personB;

PR #229, opened 2022-07-08
ZA2-TersiaKoetzee-javascript-week2

tersiakoetzee:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/B-boolean-literals/exercise.js
+++ b/exercises/B-boolean-literals/exercise.js
@@ -6,6 +6,10 @@
 */
 
 let codeYourFutureIsGreat = true;
+let mozafarIsCool = false;
+let calculationCorrect = true;
+let moreThan10Students = false;
+
 
 /* 
   DO NOT EDIT BELOW THIS LINE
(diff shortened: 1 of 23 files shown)


TemwaNyirenda (commented):
Hello Tersia, well done on your submission. You did really well for most of the exercises. Just a few to note:

For exercises/C-comparison-operators/exercise.js sameName variable, I think they wanted us to use === to check if the name are the same
 
For exercises/D-logical-operators/exercise2.js, the isBetween5and10 function, you didn't include the restriction for it being greater than or equal to 5 and less than or equal to 10.

For exercises/F-predicates/exercise.js, same issue as above, no restriction for it to be less than or equal to 10.

For exercises/G-conditionals-2/exercise-3.js, if someone got 30 they would be graded a "C" because your last else if is for anything less than or equal to 60 when it should have been between 60 and 50

PR #233, opened 2022-07-08
ZA2-Berenice-Ogundele-Javascript-Core-1-Coursework-Week2

BereniceOgundele:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/G-conditionals-2/exercise-4.js
+++ b/exercises/G-conditionals-2/exercise-4.js
@@ -8,7 +8,15 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+  let string = sentence;
+  let word = string.includes("code");
+  if ((string = word)) {
+    return "true";
+  } else {
+    return "false";
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/J-array-get-set/exercise.js
+++ b/exercises/J-array-get-set/exercise.js
@@ -6,14 +6,18 @@
 
 function first(arr) {
   return; // complete this statement
-  return arr.first(0, 1)[0];
+  return numbers[0];
 }
 
 function last(arr) {
   return; // complete this statement
-  return arr[arr.length - 1];
+  if (arr===numbers) {
+    return numbers[2];
+  }
+  else if (arr === names) {
+    return names[3];
+  }
 }
-
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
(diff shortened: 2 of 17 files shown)


exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,15 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+  let string = sentence;
+  let word = string.includes("code");
+  if ((string = word)) {
```
chandredewet, line 14: In line 14 to note that here you used the "=" Assignment operator instead of the Comparison operator which would mean that instead of comparing you are making this statement equal the value of word.  so if for e.g string includes Code, word would be true, but then String would now be reassigned true, and it would read as "If String is true - Return "true". But if string did not include code. word would hold false but string would still be assigned false and it would read as "If string is false - Return true.

exercises/J-array-get-set/exercise.js
```diff
@@ -6,12 +6,18 @@
 
 function first(arr) {
   return; // complete this statement
+  return numbers[0];
```
chandredewet, line 9: Hi Berenice,

I believe these return statements in line 8 and 14 may cause issues in your code. Once a return statement is run, it kicks out of the function or program. So it won't return line 9 or check  your conditional in line 14. if you remove line 8and 14 and always keep the return as the last statement you should be fine, or as in when you want the program to leave the function and return back to where it was called, or return a value.

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,10 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  return (typeof num === "number" && num 2 === 0 && num <= 100);
```
chandredewet, line 10: I think the num 2 === 0 should have been num 2%0 to check for even numbers? Perhaps I am getting this wrong, but I do believe that num 2 === 0 is not a valid expression. Maybe it was supposed to be num === 0

chandredewet (commented):
Hi Berenice,

Well done on amazing code, neat, legible, well indented, with just a space here and there that could be removed. I also saw that you do know the basics that can trip one up, e.g. using strict equality and your Comparison operands. There is little to correct as I see you have done all of the difficult Mandatory ones and well done to you for mapping and chaining the final one. And I didn't personally manage to get the decimal/format percentage one right, and see I can learn from using your Math,round

The only changes I would recommend is the assignment/comparison I mentioned in comments, then return statements comment and the modulus comment /num === 0 under mandatory.

Otherwise well done, I knew your work, numbers, maths brain would be in action and I can see that you have a keen understanding of javascript

PR #265, opened 2022-08-10
WestMidland 4 - Kawa manmi-JavaScript-Core1-Week2

kawaMANMI:

# Your Details

- Your Name: Kawa Manmi
- Your City: WestMidland(Birmingham)
- Your Slack Name: Kawa Manmi

# Homework Details

- Module: JavaScript-Core1
- Week: 2

# Notes

- What did you find easy?  Moderate to me

- What did you find hard? Nothing Actually

- What do you still not understand?N/A

- Any other notes?

### DIFF
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -7,7 +7,7 @@ Tip: use logical operators
 */
 
 function validate(num) {
-if (typeof num ===Number && num % 2===0 && num <= 100)
+if  ((typeof num ==="number") && (num % 2===0) && (num <= 100))
 {
   return true
 }
@@ -26,7 +26,7 @@ Write a function that:
 */
 
 function formatPercentage(num) {
-let newNum=num.toFixed(2);
+let newNum=Math.round(num*100)/100;
 if (newNum > 100) newNum=100;
 return newNum+"%";
 }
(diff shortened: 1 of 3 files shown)


kawaMANMI:
@jdbevan

mandatory/2-function-creation.js
```diff
@@ -16,17 +25,28 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
-
+function formatPercentage(num) {
+let newNum=Math.round(num*100)/100;
+if (newNum > 100) newNum=100;
```
AzinYad, line 30: I would right this line of could before Math.round one, so if the number is bigger than 100 it wont go trough the whole process and just return 100%, I think it might lead the process time to be shorter and executed quicker.

mandatory/2-function-creation.js
```diff
@@ -16,17 +25,28 @@ Write a function that:
...
 - takes an array of strings as input
 - removes any spaces in the beginning or end of each string
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
-
+function tidyUpStrings(arrayOfStrings) {
+  for (let i=0; i<arrayOfStrings.length; i++){
+    arrayOfStrings[i]= arrayOfStrings[i].trimStart();
```
AzinYad, line 42: It's not wrong but i suggest using .trim() function would trim both start and end side of string, and will save you from writing one more line of code  :)

mandatory/3-playing-computer.js
```diff
@@ -6,13 +6,17 @@
...
+  1. This program throws an error. Why? 
+  (If you can't find it, try executing it).
+  Ans: line start with console.log(b); as b is not defined
+  2. Remove the line that throws the error. Ans: Removed
+  3. What is printed to the console? 
+  Ans :value(s) of x, a, f2(0, 2)=4, f1(1,8)=9, f2(2,2)=6, f1(3,10)=13
+  f2(4,2)=8.
+  4. How many times is "f1" called?Ans: 2
+  5. How many times is "f2" called?Ans: 3
+  6. What value does the "a" parameter take in the first "f1" call? 1
```
simeonbikov, line 18: I think it is 8

mandatory/2-function-creation.js
simeonbikov: Azin is right

simeonbikov (commented):
Everything looks good

mandatory/2-function-creation.js
jdbevan: This is a good thing to be thinking about. How to "return early" from a function to avoid doing unecessary work.

mandatory/2-function-creation.js
jdbevan: Instead of accessing and then overwriting the `arrayOfStrings[i]` value on each line, can you think of a way to simplify this?

PR #271, opened 2022-08-11
Westmidlands-class 4-farzad-nazif- JavaScript-Core-1-Coursework-Week2

farzad-nazif:

# Your Details

- Your Name: Farzad Nazif
- Your City: Westmidlands - Birmingham
- Your Slack Name: farzad-nazif

# Homework Details

- Module: JavaScript-Core-1-Coursework
- Week: week2

# Notes

- What did you find easy?
creating functions
- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,10 +10,10 @@
 
 */
 
-function mood() {
+function mood(happy) {
   let isHappy = true;
 
-  if (isHappy) {
+  if (isHappy === happy) {
     return "I am happy";
   } else {
     return "I am not happy";
@@ -21,9 +21,9 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough=10;
 
-  if (isBigEnough) {
+  if (isBigEnough<num) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
@@ -32,7 +32,7 @@ function greaterThan10(num) {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element=arr[index];
 
   return element;
 }
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -6,7 +6,9 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  return typeof num === typeof 2 && num%2===0 && num<=100;
+}
 
 /*
 Write a function that:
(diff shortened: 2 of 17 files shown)


mandatory/1-fix-functions.js
```diff
@@ -32,7 +32,7 @@ function greaterThan10(num) {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element=arr[index];
 
   return element;
 }
```
alicagatay, line 38: Correct.

You could also just do `return arr[index]` but your way is also acceptable.

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,9 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  return typeof num === typeof 2 && num%2===0 && num<=100;
+}
```
alicagatay, line 11: Correct.

mandatory/2-function-creation.js
```diff
@@ -16,7 +18,20 @@ Write a function that:
...
+  }
+  else if(result===num&&num<100){
+    return `${num}%`;
+  }
+  else{
+    let rounded = num.toFixed(2);
+    let result = parseFloat(rounded);
+    return `${result}%`;
+  }
+}
```
alicagatay, line 34: Again, a nice use of functions like `floor()`, 'toFixed()` and `parseFloat()`.

mandatory/3-playing-computer.js
```diff
@@ -8,11 +8,11 @@
...
   2. Remove the line that throws the error.
-  3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  3. What is printed to the console? Awnser: 2,6,4,9,6,13,8
+  4. How many times is "f1" called?  Awnser: 2 times (2 odd number 1 3)
+  5. How many times is "f2" called?  Awnser: 3 times (3 even number 0 2 4)
+  6. What value does the "a" parameter take in the first "f1" call?  Awnser: 8
```
alicagatay, line 14: Declaration of `f1`:

`
const f1 = function (a, b) {
  return a + b;
};
`

We can see from the creation of f1 that its first parameter is a. Therefore whenever f1 is called, its first argument will be assigned to the first parameter a. Let's look at where f1 is called:

`
} else {
  const e = f1(i, a);
  console.log(e);
}
`
`f1` is called when `i` is either 1 or 3.

Therefore, the first time `f1` is called `i` must be equal to 1.

So `f1`'s first argument is 1 and therefore inside the function `f1`, `a` will be assigned a value of 1

mandatory/3-playing-computer.js
```diff
@@ -8,11 +8,11 @@
...
+  7. What is the value of the "a" outer variable when "f1" is called for the first time? Awnser: 6
```
alicagatay, line 15: The outer variable `a` is initialised with a value of 6.

This value is then incremented by 1 for each iteration of the `for` loop.

Therefore, just before we call `f1` for the first time then `i` must have a value of 1: therefore `a` will have been incremented by 1 twice already. Therefore `a` will have a value of 8.

alicagatay (changes requested):
Another great work Farzad.

There was some small errors regarding the last question, but other than that everything was great. Let me know if you have any questions for me regarding my comments for the last questions.

Keep up the good work!

PR #279, opened 2022-08-12
Driss Hmaidi - JavaScript-Core-1-Coursework-Week2

ghost:
West Midlands Class 4, Driss_Hmaidi - JavaScript Week 2


alicagatay (changes requested):
Again a nice job Driss. There was a few problems towards the end, but in general your answers were great. Keep up the great work.

PR #295, opened 2022-09-13
NW5-Manchester-Mustafa Acar-JavaScript-Core-1-Coursework-Week2

MustafaAcar-sys:

# Your Details

- Your Name:Mustafa Acar
- Your City: Manchester
- Your Slack Name: MustafaAcar-sys

# Homework Details

- Module:JavaScript-Core-1
- Week:1

# Notes

- What did you find easy?

- What did you find hard?
I could not understand the last function check exercise.

- Any other notes?
@Dominic-Taylor-Dev

### DIFF
--- a/exercises/G-conditionals-2/exercise-3.js
+++ b/exercises/G-conditionals-2/exercise-3.js
@@ -8,7 +8,17 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80) {
+    return "A";
+  } else if (mark < 80 && mark > 60) {
+    return "B";
+  } else if (mark <= 60 && mark > 50) {
+    return "C";
+  } else {
+    return "F";
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/I-array-properties/exercise.js
+++ b/exercises/I-array-properties/exercise.js
@@ -6,7 +6,7 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  return arr.length===0; // complete this statement
 }
 
 /* 
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,8 +10,7 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(isHappy) {
 
   if (isHappy) {
     return "I am happy";
@@ -21,9 +20,9 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough=10;
 
-  if (isBigEnough) {
+  if (isBigEnough<num) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
@@ -31,8 +30,8 @@ function greaterThan10(num) {
 }
 
 function get3rdIndex(arr) {
-  let index = 3;
-  let element;
+  let index = 2;
+  let element=arr[index];
 
   return element;
 }
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -6,7 +6,9 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  return num <= 100 && !(num % 2) && typeof num === "number";
+}
 
 /*
 Write a function that:
@@ -16,7 +18,9 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+  let valid = num > 100 ? 100 : parseFloat(num.toFixed(2));
+return valid + "%";}
 
 /*
 Write a function that:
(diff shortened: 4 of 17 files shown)


exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,17 @@
...
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80) {
+    return "A";
+  } else if (mark < 80 && mark > 60) {
+    return "B";
+  } else if (mark <= 60 && mark > 50) {
+    return "C";
+  } else {
+    return "F";
```
Dominic-Taylor-Dev, line 19: This is correct, but you are repeating some checks that are redundant. For example, on line 12 you already check `if (mark >= 80)`. This means that on line 14 you don't need to check if `mark < 80`. It _must_  be, because if it weren't less than 80, then the code on line 13 would have executed, and you'd never have reached line 14.

So, in short, this function would do exactly the same thing:

```
  if (mark >= 80) {
    return "A";
  } else if (mark > 60) {
    return "B";
  } else if (mark > 50) {
    return "C";
  } else {
    return "F";
  }
}
```

exercises/I-array-properties/exercise.js
```diff
@@ -6,7 +6,7 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  return arr.length===0; // complete this statement
```
Dominic-Taylor-Dev, line 9: Yep, this is how I would do it too. An alternative is to write `!arr.length` - can you understand why that would also work? NOTE: I'm not suggesting that the alternative is better - it isn't. This is just to get you thinking :)

mandatory/1-fix-functions.js
```diff
@@ -10,8 +10,7 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(isHappy) {
```
Dominic-Taylor-Dev, line 13: This works. As a little tip, it's often more readable not to use `else` in a simple 'yes/no' situation like this. So in this case you could write:

```
function mood(isHappy) {

  if (isHappy) {
    return "I am happy";
  } 
  
  return "I am not happy";
}
```

That would work exactly the same because `isHappy` is either true or false. If it's true, then we return out of the function early, so `return "I am not happy"` never executes

mandatory/1-fix-functions.js
```diff
@@ -21,18 +20,18 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough=10;
 
-  if (isBigEnough) {
+  if (isBigEnough<num) {
```
Dominic-Taylor-Dev, line 25: same principle

mandatory/1-fix-functions.js
```diff
@@ -21,18 +20,18 @@ function mood() {
...
     return "num is greater than 10";
   } else {
     return "num is not big enough";
   }
 }
 
 function get3rdIndex(arr) {
-  let index = 3;
-  let element;
+  let index = 2;
```
Dominic-Taylor-Dev, line 33: From looking at the test cases at the bottom of the file, this is expecting `arr[3]` rather than `arr[2]` - you have changed the index to 2, which will give you the third element (I can see your logic!) but that isn't what the test is expecting to see.

mandatory/2-function-creation.js
```diff
@@ -16,7 +18,9 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+  let valid = num > 100 ? 100 : parseFloat(num.toFixed(2));
```
Dominic-Taylor-Dev, line 22: This is an interesting approach. I think your ternary is sensible, effectively defaulting to 100 if the number is too big, even though I don't think you covered this syntax yet in class.

mandatory/2-function-creation.js
```diff
@@ -25,7 +29,9 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+  return arrayOfStrings.map((x) => x.trim().toLowerCase().replace(/\//g, ""));
```
Dominic-Taylor-Dev, line 33: Try to give variables meaningful names e.g. in this case `x` could be `string`. Otherwise great. 

You used regex here (`/\//g`) which is perfectly fine. You could have just have a straightforward string if you'd preferred: `.replace('/', '')`

mandatory/3-playing-computer.js
```diff
@@ -17,6 +17,7 @@
 
 let x = 2;
 let a = 6;
+let b = 4;
```
Dominic-Taylor-Dev, line 20: Yes this lets the code execute by resolving the reference error. The alternative would be to remove the `console.log(b);` line, since variable `b` is not used anywhere in the code. But what you've done is sensible too - and the question doesn't tell you what the code is supposed to do, so your approach also makes sense.

Dominic-Taylor-Dev (commented):
Good job Mustafa :)

mandatory/2-function-creation.js
Dominic-Taylor-Dev: (^ but that would only replace one /)

mandatory/1-fix-functions.js
Dominic-Taylor-Dev: (you can also one-line solve this if you don't set up named variables for things like `index` and `element`; in this particular case, since the method name is clearly specifically going to target the 3rd index, and it's not a value that will ever vary, I'd consider something like:

```
function get3rdIndex(arr) { 
    return arr[3]
}
```
)

PR #321, opened 2022-09-16
LEEDS CLASS_NW5 - Shima  Adnan - JavaScript Core 1 - WEEK_2

Shimaadnan:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,8 +5,34 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
-
+function isNegative(num) {
+  if (num < 0) {
+    return true;
+  } else {
+    return false;
+  }
+}
+function isBetween5and10(num1) {
+  if (num1 < 10 && num1 > 5) {
+    return true;
+  } else {
+    return false;
+  }
+}
+function isShortName(str) {
+  if (str === "Daniel") {
+    return true;
+  } else {
+    return false;
+  }
+}
+function startsWithD(str1) {
+  if (str1[0] === "D") {
+    return true;
+  } else {
+    return false;
+  }
+}
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
--- a/exercises/G-conditionals-2/exercise-3.js
+++ b/exercises/G-conditionals-2/exercise-3.js
@@ -8,7 +8,17 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80) {
+    return "A";
+  } else if (mark > 60 && mark < 80) {
+    return "B";
+  } else if (mark < 60 && mark > 50) {
+    return "C";
+  } else {
+    return "F";
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-4.js
+++ b/exercises/G-conditionals-2/exercise-4.js
@@ -8,7 +8,13 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+  if (sentence.includes("code")) {
+    return "true";
+  } else {
+    return false;
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,8 +10,8 @@
 
 */
 
-function mood() {
-  let isHappy = true;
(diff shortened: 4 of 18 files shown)


exercises/D-logical-operators/exercise2.js
```diff
@@ -5,8 +5,34 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
-
+function isNegative(num) {
+  if (num < 0) {
+    return true;
+  } else {
+    return false;
```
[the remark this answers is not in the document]
Gevie: Thank you Ekrem.

With this specific function, we only need to check if a number is negative or positive, I think Ekrem you were looking at a different function 👍 

What you have done Shima works perfectly! We could simplify this by just returning the comparison, if you recall from lessons, something like `(num < 0)` will return true or false. Knowing this we can do something like this:

```javascript
function isNegative(num) {
  return (num < 0);
}
```

Since `num < 0` will be true or false we get the result we want :)

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,17 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
```
Gevie, line 11: Great attempt Shima, you did this very well.

Unfortunately, you do have a couple issues here. You start by using `mark >= 80` which means if the mark is 80 or higher.

Then after this you use `mark > 60` and `mark < 80`. Unfortunately these conditions **exclude** 60 and 80 since they don't have the equals sign.

You also don't need to check if the mark is less than 80 on line 14 because the first condition already shares that if it was 80 or higher it'd `return "A"`. So if it was 80 or 85 or 1120, the function would already return.

You can simplify your second condition by using

```javascript
} else if (mark >= 60) {
  return "B";
} else if ....
```

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,13 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+  if (sentence.includes("code")) {
```
Gevie, line 12: Awesome, great job Shima.

Like I shared earlier in the PR, you could simplify this without the if statement.

```javascript
  return (sentence.includes("code"));
}
```

mandatory/1-fix-functions.js
```diff
@@ -10,8 +10,8 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(isHappy) {
+  // let isHappy = true;
```
Gevie, line 14: You are absolutely correct to remove this line, you commented it out rather than deleting it, this is a smart thing to do when working on a problem.

Unfortunately, we discourage pushing work with comments like this once we are happy with the final solution, it's fine for work in progress but when you're ready to submit your final version we should always remove commented out code. There are great reasons for this, I'd be happy to share in our team channel, but if you'd like I can share this documentation from CYF: https://syllabus.codeyourfuture.io/guides/code-style-guide#dont-leave-lots-of-commented-out-code

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,13 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  if (typeof num === "number" && num % 2 === 0 && num <= 100) {
```
Gevie, line 10: Fantastic, your logic here to validate is perfect. This is what we wanted to see and you delivered.

mandatory/2-function-creation.js
```diff
@@ -16,7 +22,16 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+  if (typeof num === "number" && num <= 100) {
```
Gevie, line 26: This is an interesting approach, you did very well.

I wonder Shima, would it be good to check if the number is more than 100 first then return `100%` as a string?

exercises/D-logical-operators/exercise2.js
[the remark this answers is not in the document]
Ekremteke: Thank you, Stephen
You are right 🙈 I should have written this for exercises/F-predicates/exercise.js.
To review for our buddy group friends' PR was Home Work last week :)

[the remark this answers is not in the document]
Gevie: You did a fantastic job and I am grateful for you helping out :) Submitting PR reviews is a big part of being a developer, you smashed it mate 👍 😄

PR #328, opened 2022-09-19
NW5-Manchester-Khadar-M-Dagal-JS1-W2

khmdagal:


extra/1-radio-stations.js
```diff
@@ -25,11 +42,17 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
+function isRadioStation(num) {
```
Ara225, line 45: This function declaration causes an exception which breaks the otherwise good code. You can't declare two functions with the same name in the same scope, and in this case, an isRadioStation function already exists, so it'll throw a SyntaxError

mandatory/3-playing-computer.js
```diff
@@ -13,6 +13,42 @@
   5. How many times is "f2" called?
   6. What value does the "a" parameter take in the first "f1" call?
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
+
```
Ara225, line 16: Good explination

mandatory/2-function-creation.js
```diff
@@ -16,7 +22,21 @@ Write a function that:
...
 */
 
-function formatPercentage(num) {}
+// create if statement
+function formatPercentage(num) {
+  const percentageSign = "%";
+  const roundedToTwo = Math.round((num + Number.EPSILON) * 100) / 100;
+  if (num > 100) {
+    let num = 100;
+    const roundedToTwo = Math.round((num + Number.EPSILON) * 100) / 100;
```
Ara225, line 31: I love the way you though about your anwser and choose a pragmatic choice. I would have added comments here to make your logic clear and refer people to whereever you got the answer from but that's subjective

Ara225 (commented):
All very good work, few nit picks but that's all, looks like you've got the basics down

PR #332, opened 2022-11-21
London 9 - turing - susan - JavaScript Core 1 - Week 2 

hkuhku609:


mandatory/1-fix-functions.js
```diff
@@ -10,76 +10,74 @@
...
-    return "I am not happy";
+    return 'I am not happy'
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num > 10
 
   if (isBigEnough) {
```
samhoooo, line 24: This can be simplified to `if ( num > 10 )`

mandatory/1-fix-functions.js
```diff
@@ -10,76 +10,74 @@
...
   } else {
-    return "num is not big enough";
+    return 'num is not big enough'
   }
 }
 
 function get3rdIndex(arr) {
-  let index = 3;
-  let element;
+  let index = 3
```
samhoooo, line 32: Do we need the variable `index`?

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,9 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  return typeof num === 'number' && num % 2 == 0 && num <= 100
```
samhoooo, line 10: I like this simplicity! Concise and readable code

mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,54 @@
...
+  8
+
   4. How many times is "f1" called?
+  2
+
   5. How many times is "f2" called?
+  3
+
   6. What value does the "a" parameter take in the first "f1" call?
+  8
```
samhoooo, line 31: #### Declaration of `f1`

```js
const f1 = function (a, b) {
  return a + b;
};
```

We can see from the creation of `f1` that its first parameter is `a`. Therefore whenever `f1` is called, its **first argument** will be assigned to the **first parameter** `a`. Let's look at where `f1` is called:

#### Call site for `f1`

```js
} else {
  const e = f1(i, a);
  console.log(e);
}
```

`f1` is called when `i` is either `1` or `3`.

Therefore, the first time `f1` is called `i` must be equal to 1.

So `f1`'s first argument is 1 and therefore inside the function `f1`, `a` will be assigned a value of 1

extra/1-radio-stations.js
```diff
@@ -25,6 +32,18 @@
...
+function getStations() {
+  let arrLength = getAllFrequencies().length
+  // let arr = []
+  // for (i = 0; i < arrLength; i++) {
+  //   if (isRadioStation(getAllFrequencies()[i])) {
+  //     arr[arr.length] = getAllFrequencies()[i]
+  //     // arr.push(getAllFrequencies()[i])
+  //   }
+  // }
+  // return arr
```
samhoooo, line 44: It would be nicer to remove the unused code before commit

samhoooo (commented):
Great work! I left a few comments here, all minor through, looks good to me

PR #336, opened 2022-11-25
London 9 - Oleh Pysmenko - JavaScript Core 1 - Week 2 

tiritaki:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -14,7 +14,10 @@
  */
 
 // `getAllFrequencies` goes here
-
+function getAllFrequencies(){
+  let freqArray = [87,88,89,90,91,92,93,94,95,96,97,98,99,100,101,102,103,104,105,106,107,108]
+  return freqArray
+}
 /**
  * Next, let's write a function that gives us only the frequencies that are radio stations.
  * Call this function `getStations`.
@@ -25,7 +28,15 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
-
+function getStations(){
+  let realStations = []
+  for (i in getAllFrequencies()) {
+    if (isRadioStation(getAllFrequencies()[i]) == true){
+      realStations.push(getAllFrequencies()[i])
+    }
+  }
+  return realStations
+}
 /*
  * ======= TESTS - DO NOT MODIFY =======
  * Note: You are not expected to understand everything below this comment!
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,10 +10,10 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(bool) {
+  let isHappy = true; 
 
-  if (isHappy) {
+  if (isHappy==bool) {
     return "I am happy";
   } else {
     return "I am not happy";
@@ -21,9 +21,9 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough=10;
 
-  if (isBigEnough) {
+  if (isBigEnough<num) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
@@ -32,7 +32,7 @@ function greaterThan10(num) {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element=arr[index];
 
   return element;
 }
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -6,7 +6,12 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  if (typeof num === "string") {return false}
+  else if (num===100) {return true}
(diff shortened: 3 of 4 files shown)


extra/1-radio-stations.js
```diff
@@ -14,7 +14,10 @@
  */
 
 // `getAllFrequencies` goes here
-
+function getAllFrequencies(){
+  let freqArray = [87,88,89,90,91,92,93,94,95,96,97,98,99,100,101,102,103,104,105,106,107,108]
```
nuclear-pasta, line 18: Is there any way you could do this without manually writing the stations one by one? Maybe only defining the start and the end?

extra/1-radio-stations.js
```diff
@@ -25,7 +28,15 @@
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
-
+function getStations(){
+  let realStations = []
+  for (i in getAllFrequencies()) {
+    if (isRadioStation(getAllFrequencies()[i]) == true){
```
nuclear-pasta, line 34: the true in the if statement can also be omitted and be written as:
`if (isRadioStation(getAllFrequencies()[i]))`

mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,20 @@
 
 */
 
-function mood() {
-  let isHappy = true;
+function mood(bool) {
+  let isHappy = true; 
 
-  if (isHappy) {
+  if (isHappy==bool) {
```
nuclear-pasta, line 16: line 14 is redundant. You could evaluate directly the argument parameter. For example:
```
  if (bool) {
  return "I am happy";
  } else {
    return "I am not happy";
  }
  ```

File: mandatory/2-function-creation.js [hunk 4]
```diff
@@ -6,7 +6,12 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  if (typeof num === "string") {return false}
```
nuclear-pasta, line 10: What would be the output for this line if I pass an object like
```
{foo: 1}
````

File: mandatory/2-function-creation.js [hunk 5]
```diff
@@ -6,7 +6,12 @@ Complete the function to check if the variable `num` satisfies the following req
...
+  else if (num===100) {return true}
+  else if (num%2==0 & num <= 100 ) {return true} 
+  return false
```
nuclear-pasta, line 13: you could concatenate the statements with an `&&` and omit the true to make it more compact.

nuclear-pasta (commented):
nice job! Keep up the good work

extra/1-radio-stations.js
tiritaki: changed it to a loop

mandatory/1-fix-functions.js
tiritaki: yes, fixed it

mandatory/2-function-creation.js
tiritaki: false - because this is an object dictionary and not string

mandatory/2-function-creation.js
tiritaki: yes, changed it to &&

mandatory/2-function-creation.js
nuclear-pasta: and what would happen if I pass number 2
Screenshot 2022-11-28 at 09 36 21

nuclear-pasta: my point was: would it make more sense to validate for a number than a string?

mandatory/2-function-creation.js
nuclear-pasta: that's cool. Line 11 is not redundant as it is included in line 12 <**=**

nuclear-pasta: also usually when you have statement that return booleans you can write them in a compact form.
For example, if I had:
```
function(banana){
   if( banana == true) {
   return false
  }
}
```
I could just go for:
```


```
function(banana){
   return !banana
}
```
if the statement is true, it is even simpler cause I don't need the bang `!`

mandatory/2-function-creation.js
nuclear-pasta: is there any way I could pass to line 11 in the current code?

nuclear-pasta: awesome :) . First two line could even be joined together with an &&

tiritaki: optimised to one line
return typeof num === "number" && num%2==0 && num <= 100 ||  num===100 ?  true : false

nuclear-pasta: do you ned num === 100 ? that is already included in num <= 100.
you could just retun this, without even the ternary statement.
`return typeof num === "number" && num%2==0 && num <= 100`

tiritaki: thank you! Done!

PR #340, opened 2022-11-27
London 9 Turing - Zahraa Tayyar - JavaScript 2

ZahraaTayyar:

# Your Details

- Your Name: Zahraa Tayyar
- Your City: London 9
- Your Slack Name: Zahraa Tayyar

# Homework Details

- Module: JavaScript 2
- Week: 2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/B-boolean-literals/exercise.js
+++ b/exercises/B-boolean-literals/exercise.js
@@ -6,6 +6,9 @@
 */
 
 let codeYourFutureIsGreat = true;
+let mozafarIsCool = false;
+let calculationCorrect = true;
+let moreThan10Students = false;
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,15 @@
 
 let studentCount = 16;
 let mentorCount = 9;
-let moreStudentsThanMentors; // finish this statement
+let moreStudentsThanMentors = studentCount > mentorCount;
 
 let roomMaxCapacity = 25;
-let enoughSpaceInRoom; // finish this statement
+let totalStudentAndMentor = studentCount + mentorCount;
+let enoughSpaceInRoom = totalStudentAndMentor <= 25; // finish this statement
 
 let personA = "Daniel";
 let personB = "Irina";
-let sameName; // finish this statement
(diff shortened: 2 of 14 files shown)


mandatory/2-function-creation.js
```diff
@@ -6,7 +6,15 @@ Complete the function to check if the variable `num` satisfies the following req
...
-function validate(num) {}
+function validate(num) {
+  if (!isNaN(num) && num % 2 === 0 && num <= 100) {
+    return true;
+  }  else {
+    return false;
+  }
+}
+
+console.log(validate("name"));
```
SallyMcGrath, line 18: This is quite complicated! It's really common to get in a muddle like this at this point of the course, so don't despair!

There are a few small things here that might be useful to clear up:

1. An expression that evaluates to true is effectively the same as true. You can return the result of the expression directly and you don't need to also return true. Think of the expression as a door. If the expression evaluates to true then the door opens and the code contained within the block is executed. 

2. Try taking out the else block and just putting `return false` instead. What happens? Why? Why not?

mandatory/2-function-creation.js
```diff
@@ -25,7 +43,11 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {   // comment: i don't understand
+
```
SallyMcGrath, line 47: I think this function is an opportunity to practice some of the methods you have learned up to now. Let's break down the requirements:

- takes an array of strings as input => the parameter will take an array and we want to do the same operation or set of operations on each string in the array

What operations should we do?
- removes any spaces in the beginning or end of each string => is there a useful way to `trim` these strings?
- removes any forward slashes (/) in each string => Can we `replace` the forward slashes? Look in the mdn docs
- makes all strings all lowercase => what about here? Can you think of a method that could achieve this?

Once we have sorted out what we will do to each string, we need to find a way to apply this to each string programmatically. Is there a way to loop over every item in your array and apply these transformations? Look through the syllabus for this week to help you.

SallyMcGrath (commented):
Really good try @ZahraaTayyar . I really appreciated the way you made notes about the parts you didn't understand so I could quickly get help to you. 

Take a look through the comments I have made and take your questions to class. Good work and thanks again!

PR #345, opened 2022-11-28
London Class 9 - Andriana Saffo - JavaScript - Week 2

AndrianaOS:

# Your Details

- Your Name: Andriana Saffo
- Your City: London
- Your Slack Name: Andriana

# Homework Details

- Module: JavaScript
- Week: 2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-radio-stations.js
+++ b/extra/1-radio-stations.js
@@ -15,13 +15,18 @@
 
 // `getAllFrequencies` goes here
 function getAllFrequencies() {
-  return listOfFrequencies;
-}
+  const allFrequencies = [];
 
-const listOfFrequencies = [
-  87, 88, 89, 90, 91, 92, 93, 94, 95, 96, 97, 98, 99, 100, 101, 102, 103, 104,
-  105, 106, 107, 108,
-];
+  const startingFreq = 87;
+
+  const endingFreq = 108;
+
+  for (let i = startingFreq; i <= endingFreq; i++) {
+    allFrequencies.push(i);
+  }
+
+  return allFrequencies;
+}
 
 /**
  * Next, let's write a function that gives us only the frequencies that are radio stations.
@@ -33,6 +38,19 @@ const listOfFrequencies = [
  * - Return only the frequencies that are radio stations.
  */
 // `getStations` goes here
+
+function getStations() {
+  let allFrequencies = getAllFrequencies();
+
+  let radioStations = [];
+
+  for (let i = 0; i < allFrequencies.length; i++) {
+    if (isRadioStation(allFrequencies[i])) {
+      radioStations.push(allFrequencies[i]);
+    }
+  }
+  return radioStations;
+}
 
 /*
  * ======= TESTS - DO NOT MODIFY =======


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 let studentCount = 16;
 let mentorCount = 9;
-let moreStudentsThanMentors; // finish this statement
+let moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 let roomMaxCapacity = 25;
-let enoughSpaceInRoom; // finish this statement
+let enoughSpaceInRoom = roomMaxCapacity === studentCount + mentorCount; // finish this statement
```
gladwindos, line 13: This is good and works, but to help readability it's good to put calculations that go together in brackets like so:
```
let enoughSpaceInRoom = roomMaxCapacity === (studentCount + mentorCount); 
```

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,38 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(number) {
+  if (number <= 0) {
```
gladwindos, line 9: Since 0 is not negative (or positive) it shouldn't pass as true. So this line should be `if (number < 0)`

exercises/E-conditionals/exercise.js
```diff
@@ -9,6 +9,12 @@
 let name = "Daniel";
 let danielsRole = "mentor";
 
+if (danielsRole = "mentor") && (name = "Daniel") {
```
gladwindos, line 12: I think you meant to do:
```
if (danielsRole === "mentor") && (name === "Daniel")
```

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,7 +5,38 @@
...
+  return false;
+}
+
+function isBetween5and10(num) {
+  if (num >= 5 && num <= 10) {
+    return true;
+  }
+
+  return false;
+}
```
gladwindos, line 22: Any function that has simple check like this one, you can just return the comparison instead like so:

```
function isBetween5and10(num) {
  return num >= 5 && num <= 10;
}
```

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,17 @@
...
+  if (mark >= 80) {
+    return "A";
+  } else if (mark < 80 && mark > 60) {
+    return "B";
+  } else if (mark <= 60 && mark >= 50) {
+    return "C";
+  }
+
+  return "F";
+}
```
gladwindos, line 21: In this case, you can actually just do:
```
  if (mark >= 80) {
    return "A";
  } else if (mark >= 60) {
    return "B";
  } else if (mark >= 50) {
    return "C";
  }

  return "F";
}
```
You don't need to check `mark < 80` in the second if statement because if it has reached that line, we definiely know that it is less than 80 or it would have passed in the first if statement. So in the second if statement, we only need to check `mark > 60`. I hope this makes sense?

gladwindos (commented):
This looks really good, well done! I have added a few comments for you to have a look at.

PR #346, opened 2022-11-28
London 9 - Lovelace - Zobeir - JavaScript-Core-1 - Week 2

Zobeir-Rigi:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/B-boolean-literals/exercise.js
+++ b/exercises/B-boolean-literals/exercise.js
@@ -6,7 +6,9 @@
 */
 
 let codeYourFutureIsGreat = true;
-
+let mozafarIsCool = false;
+let calculationCorrect = true;
+let moreThan10Students = false;
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 let studentCount = 16;
 let mentorCount = 9;
-let moreStudentsThanMentors; // finish this statement
+let moreStudentsThanMentors = true ; // finish this statement
 
 let roomMaxCapacity = 25;
-let enoughSpaceInRoom; // finish this statement
+let enoughSpaceInRoom = true ; // finish this statement
 
 let personA = "Daniel";
 let personB = "Irina";
-let sameName; // finish this statement
+let sameName = false; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE


mandatory/2-function-creation.js
```diff
@@ -16,7 +18,10 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+    
+      (num > 100)?   num.replace("100")   : `% ${num.toFixed(2)} `;
+}
```
lukedsouza99, line 24: Really good to see the use of a ternary operator, that's quite advanced! The only things missing here is adding the % sign to numbers greater than 100, and a return statement, but apart from that this is good!

mandatory/2-function-creation.js
```diff
@@ -25,7 +30,11 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+     return arrayOfStrings.trim().replace("/","") .toLowercase() ;
```
lukedsouza99, line 34: Great start to this one, this line of code is completely correct! There are a couple more bits to add, have a read of the question again and see what the input on the function is (is it just one string?)

mandatory/3-playing-computer.js
```diff
@@ -9,10 +9,10 @@
   1. This program throws an error. Why? (If you can't find it, try executing it).
   2. Remove the line that throws the error.
   3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  4. How many times is "f1" called? 2 time
```
lukedsouza99, line 12: Correct!

mandatory/3-playing-computer.js
```diff
@@ -9,10 +9,10 @@
...
+  5. How many times is "f2" called? 2 time
```
lukedsouza99, line 13: Try having another go at counting this

mandatory/3-playing-computer.js
```diff
@@ -9,10 +9,10 @@
...
+  6. What value does the "a" parameter take in the first "f1" call? 7
+  7. What is the value of the "a" outer variable when "f1" is called for the first time? 7
```
lukedsouza99, line 15: Correct!

mandatory/3-playing-computer.js
```diff
@@ -9,10 +9,10 @@
   1. This program throws an error. Why? (If you can't find it, try executing it).
   2. Remove the line that throws the error.
   3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  4. How many times is "f1" called? 2 time
+  5. How many times is "f2" called? 2 time
+  6. What value does the "a" parameter take in the first "f1" call? 7
```
lukedsouza99, line 14: This isn't a very well worded question, it  confused me at first too! I think it's asking for what value is put into the function i.e. looking at function f2:

const f1 = function (a, b) {
  return a + b;
};

it wants to know what value is being put in the a parameter. Like I said, it's not the easiest question and wasn't written very well I don't think

mandatory/2-function-creation.js
Zobeir-Rigi: why should I add a percent sign here ?

mandatory/2-function-creation.js
Zobeir-Rigi: can you explain it more? I really dont know what you meant and what shall I add .

mandatory/3-playing-computer.js
Zobeir-Rigi: I think only two times , when i=2 and i=4 .because the condition is i<5 so it is not going for another round. can explain it please.

mandatory/3-playing-computer.js
Zobeir-Rigi: you know when i ==1 we can call f1 so its 7
im not sure

mandatory/2-function-creation.js
lukedsouza99: Hey Zobeir, this was just so that it would return "100%" rather than "100". Purely a formatting thing!

mandatory/3-playing-computer.js
lukedsouza99: Hey, there's also i = 0 as well!

mandatory/3-playing-computer.js
lukedsouza99: Hey, yes so for this question the answer is 1 because i == 1 and acts as the parameter where a is in f1

PR #351, opened 2022-11-29
London 9 - Marziyeh Azhdari - JavaScript-Core1 - Week 2

MariAzhdari:

# Your Details

- Your Name:Marziyeh
- Your City:London
- Your Slack Name:MariAzh

# Homework Details

- Module:JS
- Week:2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/F-predicates/exercise.js
+++ b/exercises/F-predicates/exercise.js
@@ -6,10 +6,22 @@
 */
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  if(number < 0){
+    return true;
+  }else{
+    return false;
+  }
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  if(number < 0){
+    return false;
+  }else{
+    return true;
+  }
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-2.js
+++ b/exercises/G-conditionals-2/exercise-2.js
@@ -7,7 +7,13 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if(grade < 50 ){
+    return "failed"
+  }else if(grade > 50){
+    return "passed"
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-3.js
+++ b/exercises/G-conditionals-2/exercise-3.js
@@ -8,7 +8,18 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if(mark >= 80){
+    return "A"; 
+  }else if(mark < 80 && mark > 60){
+    return "B";
+  }else if (mark <= 60 && mark >= 50){
+    return "C";
+  }else{
+    return "F"
+  }
+  
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-4.js
+++ b/exercises/G-conditionals-2/exercise-4.js
@@ -8,7 +8,13 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) { 
+  if (sentence .includes("code")){
+    return true;
+  }else{
+     return false;
+  }
+}
 
 /* 
 DO NOT EDIT BELOW THIS LINE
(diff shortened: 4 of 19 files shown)


exercises/F-predicates/exercise.js
```diff
@@ -6,10 +6,22 @@
...
+    return true;
+  }else{
+    return false;
+  }
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  if(number < 0){
```
mcarballopacheco, line 19: This should calcuate if the number is between 0 and 10. This function would not work for 20, it will return true but it should be false.

exercises/G-conditionals-2/exercise-2.js
```diff
@@ -7,7 +7,13 @@
 
 */
 
-function studentPassed(grade) {}
+function studentPassed(grade) {
+  if(grade < 50 ){
+    return "failed"
+  }else if(grade > 50){
```
mcarballopacheco, line 13: It should be >=, because 50 is also a passing grade.

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,18 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if(mark >= 80){
+    return "A"; 
+  }else if(mark < 80 && mark > 60){
```
mcarballopacheco, line 14: Here, you don't need the first condition: `mark < 80`. It will only pass into this else if condition if it is already smaller than 80 (becuse it has to fail the first if statement).

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,18 @@
...
+    return "B";
+  }else if (mark <= 60 && mark >= 50){
```
mcarballopacheco, line 16: The same as before here. You don't need the `mark <= 60`

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,13 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) { 
+  if (sentence .includes("code")){
```
mcarballopacheco, line 12: There should not be a whitespace between sentence.includes

exercises/I-array-properties/exercise.js
```diff
@@ -6,8 +6,13 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  if(arr.length){
```
mcarballopacheco, line 9: This would not work. The arr has a length even if it is empty. However, it will be equal to 0, so you should compare it to 0.

mandatory/1-fix-functions.js
```diff
@@ -10,19 +10,34 @@
...
+//   let isBigEnough;
 
+//   if (isBigEnough) {
+//     return "num is greater than 10";
+//   } else {
+//     return "num is not big enough";
+//   }
+// }
+function greaterThan10(num) {
+  let isBigEnough = num > 10;
```
mcarballopacheco, line 40: Great!

mandatory/2-function-creation.js
```diff
@@ -25,7 +52,11 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+ return arrayOfStrings.map((element) =>
+    element.trim().replace('/', '').toLowerCase()
+  )
```
mcarballopacheco, line 58: Great use of method chaining!

mandatory/3-playing-computer.js
```diff
@@ -6,13 +6,13 @@
...
-  3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  1. This program throws an error. Why?Because b is  undefined.(If you can't find it, try executing it).
+  2. Remove the line that throws the error.I did.
+  3. What is printed to the console? 
+  4. How many times is "f1" called? 2 times
+  5. How many times is "f2" called? 3 time but I am not sure and I have question.
```
mcarballopacheco, line 13: 3 is correct. What is the question?

mandatory/3-playing-computer.js
```diff
@@ -6,13 +6,13 @@
...
+  6. What value does the "a" parameter take in the first "f1" call?6
```
mcarballopacheco, line 14: it would be 1. a in the f1 is equal to the i outer variable. In the first iteration i = 0, but f2 does not get called. But in the second iteration f2 = 1 and then it gets called.

mcarballopacheco (commented):
Great work Mari!

PR #354, opened 2022-11-30
London Class 9 - Farzaneh Haghani - JS - Week 2

farzaneh-haghani:
# Your Details

- Your Name: Farzaneh Haghani
- Your City: London
- Your Slack Name:  Farzaneh Haghani

# Homework Details

- Module: JS
- Week:3 

# Notes

- What did you find easy? every thing

- What did you find hard? nothing

- What do you still not understand? So far, nothing 

- Any other notes? No

### DIFF
--- a/exercises/B-boolean-literals/exercise.js
+++ b/exercises/B-boolean-literals/exercise.js
@@ -6,6 +6,9 @@
 */
 
 let codeYourFutureIsGreat = true;
+let mozafarIsCool = false;
+let calculationCorrect = true;
+let moreThan10Students = false;
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,8 +5,21 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(num) {
+  return num < 0;
+}
 
+function isBetween5and10(num) {
+  return num > 5 && num < 10;
+}
+
+function isShortName(str) {
+  return str.split(" ").length === 1;
+}
+
+function startsWithD(str) {
+  return str[0] === "D";
+}
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -9,9 +9,9 @@
 let name = "Daniel";
 let danielsRole = "mentor";
 if (danielsRole === "mentor")
-  return `Hi, I 'm ${name}, I am a mentor`;
+  console.log(`Hi, I'm ${name}, I'm a mentor`);
 else
-  return `Hi, I 'm ${name}, I am a student`;
+  console.log(`Hi, I'm ${name}, I'm a student`);
 // We could solve this with only one return instead of using if.
 
 /* 
--- a/exercises/E-conditionals/exercise.js
+++ b/exercises/E-conditionals/exercise.js
@@ -8,6 +8,11 @@
 
 let name = "Daniel";
 let danielsRole = "mentor";
+if (danielsRole === "mentor")
+  return `Hi, I 'm ${name}, I am a mentor`;
+else
+  return `Hi, I 'm ${name}, I am a student`;
+// We could solve this with only one return instead of using if.
 
 /* 
 EXPECTED RESULT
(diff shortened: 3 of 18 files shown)


farzaneh-haghani:
@JDysiewicz

exercises/B-boolean-literals/exercise.js
```diff
@@ -6,6 +6,9 @@
 */
 
 let codeYourFutureIsGreat = true;
+let mozafarIsCool = false;
```
JDysiewicz, line 9: All looks good, though try to use `const` instead of `let` in the first instance, in general reassignment of variables is confusing in large codebases

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,8 +5,21 @@
...
+function isBetween5and10(num) {
+  return num > 5 && num < 10;
+}
+
+function isShortName(str) {
+  return str.split(" ").length === 1;
+}
+
+function startsWithD(str) {
+  return str[0] === "D";
```
JDysiewicz, line 21: Would do you think would happen here if an empty string was passed to this function (i.e. `startsWithD("")`)? How could you account for this?

exercises/E-conditionals/exercise.js
```diff
@@ -8,6 +8,11 @@
 
 let name = "Daniel";
 let danielsRole = "mentor";
+if (danielsRole === "mentor")
```
JDysiewicz, line 11: Make sure to use curly braces to denote code blocks after `if/else` e.g. 
```
if (X) {
...
}
```

As otherwise JS will only execute the first like of code after the statement (which is why it works here, as you're only executing 1 line)

exercises/G-conditionals-2/exercise-3.js
```diff
@@ -8,7 +8,16 @@
   - Otherwise the grade is "F"
 */
 
-function calculateGrade(mark) {}
+function calculateGrade(mark) {
+  if (mark >= 80)
+    return "A";
+  else if (mark < 80 && mark > 60)
```
JDysiewicz, line 14: The `mark < 80` check here is redundant, as any `mark > 80` will be caught in the above conditional, and `return`ed before reaching this line.

You could also use a `switch` statement to clean this up - what might that look like?

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,7 +8,11 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
+function containsCode(sentence) {
+  if (sentence.split("code").length > 1)
```
JDysiewicz, line 12: This would work, however it's a little unclear what the intent is. `split` is used to turn a string to an array based on some specifier, what you're looking for is the `.includes` method

exercises/I-array-properties/exercise.js
```diff
@@ -6,7 +6,7 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  return arr.length === 0; // complete this statement
```
JDysiewicz, line 9: Good use of returning a statement to return a Boolean value; this is a better pattern to use than something like 
```
If (arr.length === 0) {
   return true
} else {
   return false
}
```

Good job

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
```
JDysiewicz, line 8: Would what happen if an empty array was passed in?

JDysiewicz (commented):
Looks really good - the only comments I've really made are questions to get you thinking about the next steps and how this code might be used in the future. That's definitely not expected at this stage though, so don't worry! 👍

exercises/E-conditionals/exercise.js
farzaneh-haghani: Ok. Thank you. I will try to use but in Iran they taught me to use curly braces if you have more than one statement even for loop

exercises/G-conditionals-2/exercise-4.js
farzaneh-haghani: yes. I know I was dumb in this question :)

exercises/B-boolean-literals/exercise.js
farzaneh-haghani: Thank you, done.

exercises/D-logical-operators/exercise2.js
farzaneh-haghani: Thank you, done.

exercises/G-conditionals-2/exercise-3.js
farzaneh-haghani: Thank you, done. 👍

PR #358, opened 2022-11-30
London 9-Sana Asaf-Javascript-Core-1-Coursework-Week2

SanaAsaf:
Working on the extra exercise.

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -6,7 +6,14 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  if (typeof num === "number" && num %2 === 0 && num <= 100)
+  {
+    return true;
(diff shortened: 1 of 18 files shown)


mandatory/2-function-creation.js
```diff
@@ -6,17 +6,33 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  if (typeof num === "number" && num %2 === 0 && num <= 100)
```
Zobeir-Rigi, line 10: That's great. We also can use another way.
return (typeof num == "number"  &&  num%2 === 0  &&  num <= 100 );
it will be return true if all 3 condition are true  ;

PR #370, opened 2022-12-02
London 9 - Ying Xing - JavaScript-Core-1-Coursework-Week2

LexiisYing:

# Your Details

- Your Name: Ying Xing
- Your City: London
- Your Slack Name: Lexi

# Homework Details

- Module:1
- Week:2

# Notes

- What did you find easy?
not much easier
- What did you find hard?
complex function
complex functions with math involved
- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -11,7 +11,6 @@
 */
 
 function mood(isHappy) {
-  let isHappy = true;
 
   if (isHappy) {
     return "I am happy";
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,7 +10,7 @@
 
 */
 
-function mood() {
+function mood(isHappy) {
   let isHappy = true;
 
   if (isHappy) {
@@ -21,7 +21,7 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num > 10;
 
   if (isBigEnough) {
     return "num is greater than 10";
@@ -32,7 +32,7 @@ function greaterThan10(num) {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element = arr[index];
 
   return element;
 }
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -6,7 +6,9 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  return typeof num == "number" && num % 2 == 0 && num <= 100;
+}
 
 /*
 Write a function that:
@@ -16,7 +18,9 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+  return num <= 100 ? `${parseFloat(num.toFixed(2))}%` : `100%`;
+}
 
 /*
 Write a function that:
@@ -25,7 +29,10 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+  return arrayOfStrings.map(string => string.trim().replace("/" , "").toLowerCase())
+}
+
 
 /* ======= TESTS - DO NOT MODIFY ===== */
 
--- a/mandatory/3-playing-computer.js
+++ b/mandatory/3-playing-computer.js
@@ -15,7 +15,7 @@
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
 */
 
-let x = 2;
+
 let a = 6;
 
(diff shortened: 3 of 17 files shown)


mandatory/1-fix-functions.js
```diff
@@ -21,7 +20,7 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num > 10;
```
lukedsouza99, line 23: Looks good! Another way to do it would be to remove the isBigEnough variable and just have "if (num > 10)" but functionally it's identical so it only depends on what you think is more clear to other people!

mandatory/1-fix-functions.js
```diff
@@ -32,7 +31,7 @@ function greaterThan10(num) {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element = arr[index];
```
lukedsouza99, line 34: Great! Hopefully indexing starting at 0 isn't that confusing for you, I know it took me a while to get used to it!

mandatory/2-function-creation.js
```diff
@@ -6,7 +6,9 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  return typeof num == "number" && num % 2 == 0 && num <= 100;
```
lukedsouza99, line 10: Good use of the remainder symbol!

mandatory/2-function-creation.js
```diff
@@ -16,7 +18,9 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(num) {}
+function formatPercentage(num) {
+  return num <= 100 ? `${parseFloat(num.toFixed(2))}%` : `100%`;
```
lukedsouza99, line 22: Great! Good use of the ternary operator as well, that's fairly advanced!

mandatory/2-function-creation.js
```diff
@@ -25,7 +29,10 @@ Write a function that:
 - removes any forward slashes (/) in each string
 - makes all strings all lowercase
 */
-function tidyUpStrings(arrayOfStrings) {}
+function tidyUpStrings(arrayOfStrings) {
+  return arrayOfStrings.map(string => string.trim().replace("/" , "").toLowerCase())
```
lukedsouza99, line 33: Really good use of the map function! Another way would be to use the for loop, but something like the map function does the same things, uses less lines and is easier to read!

mandatory/3-playing-computer.js
```diff
@@ -40,3 +40,18 @@ for (let i = 0; i < 5; ++i) {
     console.log(e);
   }
 }
+// 1. This program throws an error. Why? (If you can't find it, try executing it).
+// x is not defined
```
lukedsouza99, line 44: For questions 1 and 2, line 18 is actually fine as is. You're on the right tracks with something not being defined though! There's another line out there trying to use something that doesn't exist

mandatory/3-playing-computer.js
```diff
@@ -40,3 +40,18 @@ for (let i = 0; i < 5; ++i) {
...
+// 2. Remove the line that throws the error.
+// line 18
+// 3. What is printed to the console?
+//
+// 4. How many times is "f1" called?
+// 2
```
lukedsouza99, line 50: Correct!

mandatory/3-playing-computer.js
```diff
@@ -40,3 +40,18 @@ for (let i = 0; i < 5; ++i) {
...
+// 5. How many times is "f2" called? 
+// 2
```
lukedsouza99, line 52: This isn't quite right, have another look at the loop and double check how many times it loops, remembering that it starts at 0

mandatory/3-playing-computer.js
```diff
@@ -40,3 +40,18 @@ for (let i = 0; i < 5; ++i) {
...
+// 6. What value does the "a" parameter take in the first "f1" call?
+// 7
+// 7. What is the value of the "a" outer variable when "f1" is called for the first time?
+// 7
```
lukedsouza99, line 56: This one is just a bit off, have another look at this but I think it may be related to starting the for loop at 1 instead of 0.

mandatory/3-playing-computer.js
```diff
@@ -40,3 +40,18 @@ for (let i = 0; i < 5; ++i) {
...
+// 2. Remove the line that throws the error.
+// line 18
+// 3. What is printed to the console?
+//
+// 4. How many times is "f1" called?
+// 2
+// 5. How many times is "f2" called? 
+// 2
+// 6. What value does the "a" parameter take in the first "f1" call?
+// 7
```
lukedsouza99, line 54: This is a really bad question in my opinion that's main purpose seems to be to be confusing. Basically what it's asking is, if we look at f1:
const f1 = function (a, b) {
  return a + b;
};

What value is being put in as the first parameter of this function when it reaches it the first time. So it wants to know what the a value in this function is rather than the "global" a value that you suggested.

Like I said, it's a very annoying question, I don't really like it that much so don't worry about this one too much

PR #380, opened 2022-12-13
 London Class9 -GAYLE THOMPSON-IGWEBIKE- JavaScript-Core-1-Coursework-Week2

GayleNThompsonIgwebike:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/D-logical-operators/exercise2.js
+++ b/exercises/D-logical-operators/exercise2.js
@@ -5,8 +5,35 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(num) {
+if (num < 0){
+  return true
+}else {
+  return false;
+}
+}
+function isBetween5and10(num){
+if (num >= 5 && num <= 10){
+ return true;
+} else {
+  return false;
+}
+}
+ function isShortName(name){
+ if (name.length <= 7){
+  return true;
+ }else {
+  return false;
+ }
+ }
+ function startsWithD(dName){
+  if (dName[0] === "D"){
+    return true;
+  }else {
+    return "Name does not start with D";
+  }
 
+ }
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
--- a/exercises/F-predicates/exercise.js
+++ b/exercises/F-predicates/exercise.js
@@ -4,12 +4,27 @@
   Write two predicate functions
   The variables should have values that match the expected results.
 */
+function isNegative(number) {
+  return typeof number === "number";
+}
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  if (number === 5) {
+    return false;
+  } else if (number === -5) {
+    return true;
+  }
+}
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  if (number.isbetween === 0 || 10) {
+    return true;
+  } else if(number.isbetween ===-5 && 10){
+    return false;
+  }
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/exercises/G-conditionals-2/exercise-4.js
+++ b/exercises/G-conditionals-2/exercise-4.js
@@ -8,8 +8,13 @@
   Hint: Google how to check if a string contains a word
 */
 
-function containsCode(sentence) {}
-
+function containsCode(sentence) {
+  if(sentence = sentence.includes("code")){
+    return true;
+  }else {
+    return false;
+}
+}
 /* 
 DO NOT EDIT BELOW THIS LINE
 --------------------------- */
--- a/exercises/I-array-properties/exercise.js
+++ b/exercises/I-array-properties/exercise.js
@@ -5,15 +5,20 @@
 
 */
 
-function isEmpty(arr) {
-  return; // complete this statement
+function isEmpty(names) {
+  if ((isEmpty = [])) {
+    return false;
+  } else if (isEmpty = (names)) {
+    return true;
+    // complete this statement
+  }
 }
 
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
 let numbers = [1, 2, 3];
-let names = [];
+let names = ["mark", "john", "lara"];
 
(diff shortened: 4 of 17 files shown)


exercises/D-logical-operators/exercise2.js
```diff
@@ -5,8 +5,35 @@
...
 */
 
-function isNegative() {}
+function isNegative(num) {
+if (num < 0){
+  return true
+}else {
+  return false;
+}
+}
```
gladwindos, line 14: This is good and is correct! But there is a way to write simple if/else statements in one line if you are just returning true/false:
```
function isNegative(num) {
  return num < 0;
}
```
This works because the function returns `true` if the `num < 0` or `false` if not

exercises/F-predicates/exercise.js
```diff
@@ -4,12 +4,27 @@
...
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {}
+function isBetweenZeroAnd10(number) {
+  if (number.isbetween === 0 || 10) {
+    return true;
+  } else if(number.isbetween ===-5 && 10){
+    return false;
+  }
+}
```
gladwindos, line 27: I don't think the `isbetween` method is doing what you think it is. A better way to write this function is this:
```suggestion
function isBetweenZeroAnd10(number) {
  if (number >= 0 && number <= 10) {
    return true;
  } else {
    return false;
  }
}
```

exercises/G-conditionals-2/exercise-4.js
```diff
@@ -8,8 +8,13 @@
...
 
-function containsCode(sentence) {}
-
+function containsCode(sentence) {
+  if(sentence = sentence.includes("code")){
+    return true;
+  }else {
+    return false;
+}
+}
```
gladwindos, line 17: This is mostly correct, however you don't need to add `sentence =`. You can just do:
```suggestion
function containsCode(sentence) {
  if(sentence.includes("code")){
    return true;
  }else {
    return false;
}
}
```

exercises/I-array-properties/exercise.js
```diff
@@ -5,15 +5,21 @@
...
-  return; // complete this statement
+function isEmpty() {
+  if ((isEmpty = numbers)) {
+    return false;
+  } else if (isEmpty = names) {
+    return true;
+    // complete this statement
+  }
+  console.log(isEmpty);
 }
```
gladwindos, line 16: An easy way to check if an array is empty is to check if the length of the array is zero like so:
```
function isEmpty(arr) {
  if (arr.length === 0) {
    return true;
  } else {
    return false;
  }
}
```

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return numbers(first[0, 2]); // complete this statement
```
gladwindos, line 8: For this function, you need to get the first item in the array which you can do this with the [0] index, like so:
```
  return arr[0]; // complete this statement
```

exercises/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
...
 }
 
 function last(arr) {
-  return; // complete this statement
+  return names(last("Joe")); // complete this statement
```
gladwindos, line 12: Similarly here, you can get the last item in the array by indexing the length of the array minus one:
```
  return arr[arr.length - 1]; // complete this statement
```

extra/1-radio-stations.js
```diff
@@ -13,7 +13,13 @@
  * - Should return this array to use in other functions
  */
 
-// `getAllFrequencies` goes here
+ function getAllFrequencies(){
+   return listOfFrequencies;
+ };
+
+ const listOfFrequencies = [87, 87.5, 90, 92.5, 98.5, 100, 104.5, 108];
```
gladwindos, line 20: This works, but for this function it would be better to define the variable inside the function like so:
```
 function getAllFrequencies(){
   const listOfFrequencies = [87, 87.5, 90, 92.5, 98.5, 100, 104.5, 108];
   return listOfFrequencies;
 };
```

gladwindos (commented):
This looks good, well done! I've left comments in a few places where things don't look quite right or code can be improved.

exercises/D-logical-operators/exercise2.js
GayleNThompsonIgwebike: Thank you Gladwin. I've been practising with this a lot. I think I'm beginning to get the hang of it. I was working with some classmates and we practiced this a lot. I feel more comfortable when I work the long way, get the test working,  then try to do it the shorter way. It's long winded but helps me to 'see' what i'm doing.

exercises/F-predicates/exercise.js
```diff
@@ -4,12 +4,27 @@
...
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
-function isNegative(number) {}
+function isNegative(number) {
+  if (number === 5) {
+    return false;
+  } else if (number === -5) {
+    return true;
+  }
+}
```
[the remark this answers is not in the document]
GayleNThompsonIgwebike: OK, I got this.... I tried it again and it passed now. Thank you..x

exercises/F-predicates/exercise.js
GayleNThompsonIgwebike: @gladwindos ,      With this one, because the [ => and >=  ]confuse me a little, I redid it like this: 
"return number > 0 && number < 11;" 

Is this lazy or cheating???

exercises/G-conditionals-2/exercise-4.js
GayleNThompsonIgwebike: Got you. Thanks...

exercises/I-array-properties/exercise.js
GayleNThompsonIgwebike: Great tip actually... thank you plenty...x

exercises/J-array-get-set/exercise.js
GayleNThompsonIgwebike: Ok, cool. Thank you...

PR #381, opened 2022-12-14
London 9 - Turing -Olena Reukova -JavaScript-Core-1-Coursework - Week2

OlenaReukova:
exercises A to D done

### DIFF
--- a/exercises/C-comparison-operators/exercise.js
+++ b/exercises/C-comparison-operators/exercise.js
@@ -7,14 +7,14 @@
 
 let studentCount = 16;
 let mentorCount = 9;
-let moreStudentsThanMentors; // finish this statement
+let moreStudentsThanMentors = studentCount>mentorCount; // finish this statement
 
 let roomMaxCapacity = 25;
-let enoughSpaceInRoom; // finish this statement
+let enoughSpaceInRoom = roomMaxCapacity >= 25; // finish this statement
 
 let personA = "Daniel";
 let personB = "Irina";
-let sameName; // finish this statement
+let sameName = personA === personB; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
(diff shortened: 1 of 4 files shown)


exercises/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 let studentCount = 16;
 let mentorCount = 9;
-let moreStudentsThanMentors; // finish this statement
+let moreStudentsThanMentors = studentCount>mentorCount; // finish this statement
 
 let roomMaxCapacity = 25;
-let enoughSpaceInRoom; // finish this statement
+let enoughSpaceInRoom = roomMaxCapacity >= 25; // finish this statement
```
qingwei91, line 13: Hmm, shouldnt we use student and mentor count here?

exercises/D-logical-operators/exercise2.js
```diff
@@ -5,8 +5,34 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
-
+function isNegative(num) {
```
qingwei91, line 8: This is correct! but there might be a way to make it simpler, 

hint: what does `num < 0` return?

This simplification apply to all functions that return true and false

mandatory/1-fix-functions.js
```diff
@@ -21,18 +19,18 @@ function mood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num;
```
qingwei91, line 22: while this work, I think it might be nicer if we do `let isBigEnough = num > 10`, normally if a variable name read like `isXXX` it indicates it is a boolean value

mandatory/1-fix-functions.js
```diff
@@ -21,18 +19,18 @@ function mood() {
...
   } else {
     return "num is not big enough";
   }
 }
 
 function get3rdIndex(arr) {
-  let index = 3;
-  let element;
+  let index = arr[3];
+  let element = index;
```
qingwei91, line 33: here I think it is nicer if we do `let element = arr[index]` and don't change index.

Your approach works too

qingwei91 (approved):
Well done

PR #399, opened 2023-02-23
Glasgow 6 - Malkit Benning  - JavaScript Core 1 - Week 2

malkitbenning:
# Your Details

- Your Name: Malkit Benning
- Your City: Glasgow
- Your Slack Name: Malkit Benning

# Homework Details

- Module: JavaScript / Core 1
- Week: 2

# Notes

- What did you find easy? working with functions

- What did you find hard? remembering syntax by memory

- What do you still not understand? Best practices for writing code

- Any other notes?

### DIFF
--- a/extra/1-factorial.js
+++ b/extra/1-factorial.js
@@ -9,7 +9,18 @@
 */
 
 function factorial(input) {
-  // TODO
+  if (input == 0) {
+    return 0;
+  } else {
+    const maxLoops = input;
+    let currentLoop = 1;
+    let total = 1;
+    while (currentLoop <= maxLoops) {
+      total = total * currentLoop;
+      currentLoop++;
+    }
+    return total;
+  }
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,14 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  const acceptableAge = 18;
+  if (userAge >= acceptableAge && isLoggedIn) {
+    return true;
+  } else {
+    return false;
+  }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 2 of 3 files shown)


mandatory/2-function-creation.js
```diff
@@ -45,7 +84,13 @@ function canRegister(age) {}
   )
 */
 
-function countReverse(number) {}
+function countReverse(number) {
+  const minNum = 1;
```
msimmdev, line 88: Suggestion: create a new variable that you can loop over and decrement rather than using the input `number`. e.g.
```
let currentNumber = number;
while (currentNumber  >= minNum) {
  console.log(number);
  currentNumber--;
}
```
The reason for this is you still have access to the unmodified user input in case you need to use it in  a future modification of the function.
Also you'll find in the future lessons that for certain types passed to a function modifying the input data could have an impact on other parts of the program (this isn't a problem for integers though so wouldn't be an issue here)

mandatory/2-function-creation.js
```diff
@@ -15,26 +22,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  }
+}
 
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
   */
-function printOddNumbers(limit) {}
+function printOddNumbers(limit) {
+  let i = 1;
+  while (i <= limit) {
```
msimmdev, line 42: Thought: I think this is OK, but one reading could be that it should log numbers between 1 and limit (non inclusive of limit). in that case you would want to do `while (i < limit)`

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,14 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  const acceptableAge = 18;
+  if (userAge >= acceptableAge && isLoggedIn) {
```
msimmdev, line 9: Nitpick: While this is OK, it's slightly better to do `isLoggedIn === true` as this type checks isLoggedIn to make sure it's passed into the function as a boolean.

extra/1-factorial.js
```diff
@@ -9,7 +9,18 @@
 */
 
 function factorial(input) {
-  // TODO
+  if (input == 0) {
+    return 0;
+  } else {
+    const maxLoops = input;
```
msimmdev, line 15: Issue: It's generally not correct to assign a `const` to a variable value, instead you should use a `let` here

msimmdev (commented):
Praise: Good job on these, you've managed to implement good solutions to all the exercises including the extra questions.

extra/1-factorial.js
malkitbenning: Thanks Michael, is there not a school of thought that we should use const wherever we can so that the value is not accidentally overwritten by someone else? For the run of this function and with the declaration scope being inside the function, I thought maxLoops should not be allowed to change? I'll bring this up at our next buddy meeting to get your thoughts as I'm a bit unclear on this.

PR #412, opened 2023-02-25
LONDON10-SAQIB_JAVED_ JavaScript-Core-1-Coursework-Week2

saqibjvd:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -13,19 +13,19 @@
 function getMood() {
   let isHappy = true;
 
-  if (isHappy) {
+  if (isHappy === true) {
     return "I am happy";
-  } else {
+  } else if (isHappy === false) {
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  // let isBigEnough; - i think we do not need this function //
 
-  if (isBigEnough) {
+  if (num > 10) {
     return "num is greater than 10";
-  } else {
+  } else if (num <= 10) {
     return "num is not big enough";
   }
 }
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,18 +10,18 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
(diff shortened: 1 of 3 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,20 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
+  // let isHappy = true;
 
-  if (isHappy) {
+  if (isHappy === true) {
```
SalihaPopal, line 16: Well done!
This looks logical that isHappy === true .

mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,20 @@
...
     return "I am happy";
   } else {
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num;
```
SalihaPopal, line 24: I add 10 instead of num here. You have done it opposit of mine.

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,13 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge >= 18 && isLoggedIn === true) {
```
SalihaPopal, line 8: Should we use true two times? It is a qustion.

SalihaPopal (commented):
Will done! wish you all the best.

mandatory/2-function-creation.js
saqibjvd: This === compare if value is true for example if 2 === 2

mandatory/1-fix-functions.js
saqibjvd: I added num because in argument it's(num) but i think you can add number aswell

PR #413, opened 2023-02-25
Glasgow Class 6 - Christina Mifsud - JS1 - Week 2

christina-mifsud:
# Your Details

- Your Name: Christina Mifsud
- Your City: Glasgow
- Your Slack Name: Christina Mifsud

# Homework Details

- Module: JS1
- Week: 2

# Notes

- What did you find easy?
The isAcceptableUser exercise, the buyTwoGetTheCheapestFree exercise, the canRegister exercise.

- What did you find hard?
The countReverse exercise and the printOddNumbers exercise. I still have not managed to solve them.

I cannot grasp the while loop. I keep getting into infinity loops.

- Any other notes?
It would be helpful to have the exercises split on more js files as the numbers of tests running made it hard to find and read results.

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -9,10 +9,8 @@
   NOTE:Only make edits inside the function
 
 */
-
-function getMood() {
-  let isHappy = true;
-
+let isHappy = false;
+function getMood(isHappy) {
   if (isHappy) {
     return "I am happy";
   } else {
@@ -20,10 +18,9 @@ function getMood() {
   }
 }
 
+let isBigEnough = 9;
 function greaterThan10(num) {
-  let isBigEnough;
-
-  if (isBigEnough) {
+  if (num > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,15 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+let userAge = 18;
+let isLoggedIn = false;
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge >= 18 && isLoggedIn === true){
+return true;
+  } else {
+    return false;
+  }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
@@ -15,18 +23,49 @@ function isAcceptableUser(userAge, isLoggedIn) {}
     is applieds and 142.5 should be returned)
 */
 
-function applyDiscount(totalPrice) {}
+function applyDiscount(totalPrice) {
+  let fullDiscount = (totalPrice * 10)/100;
+  let smallDiscount = (totalPrice * 5)/100;
+  if (totalPrice >= 200){
+  return newPrice = totalPrice - fullDiscount;
+  } else {
+  return newPrice = totalPrice - smallDiscount;
+  }
+}
 
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
-  */
-function printOddNumbers(limit) {}
+*/
+
+///////////////// HELPPPP!!!!!!! /////////////////////////////
+let lowNumber = 1;
+
+function printOddNumbers(limit) {
+  while (lowNumber < limit) {
+    if (lowNumber % 2 === 0){
+    console.log(lowNumber)
+    }
+  lowNumber = lowNumber + 1;
+  }
+}
+//////////////////////////////////////////////////////////////
+
 
 /*
   Complete the buyTwoGetTheCheapestFree function: if user buys two items, the cheapest item will be free!
   The function should return the price to be paid once the discount is applied.
 */
-function buyTwoGetTheCheapestFree(price1, price2) {}
+
+function buyTwoGetTheCheapestFree(price1, price2) {
+  if (price1 && price2) {
+    if (price1 < price2) {
+      return price2
+    } else {
+      return price1
+    } 
+   }
+}
+
 
 /*
   Complete the function to determine if it is suitable for a person to register based on their age!
@@ -34,7 +73,16 @@ function buyTwoGetTheCheapestFree(price1, price2) {}
   - if the person is older than 12 and younger than 90 it should return "You Can Register"
   - if the person is 90 or older it should return "You Don't Need To Register"
 */
-function canRegister(age) {}
+
+function canRegister(age) {
+  if (age <= 12) {
+    return "You Are Too Young To Register";
+  } else if (age > 12 && age < 90) {
+    return "You Can Register";
+  } else { 
+    return "You Don't Need To Register";
+  }
+}
 
 /*
   Complete the function so that it prints out to the console numbers in reverse order starting at
@@ -45,7 +93,15 @@ function canRegister(age) {}
   )
 */
 
-function countReverse(number) {}
+/////////////////////// HELP!! ////////////////////////////////////
+function countReverse(number) {
+  while (number > 0 && number <= 10) {
+    console.log(number);
+    number = number -1;
+  }
+}
+
+//////////////////////////////////////////////////////////////
 
 /* ======= TESTS - DO NOT MODIFY ===== */
 
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -37,18 +37,15 @@ function applyDiscount(totalPrice) {
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
 */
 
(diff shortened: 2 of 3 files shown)


mandatory/1-fix-functions.js
```diff
@@ -9,21 +9,18 @@
...
-function getMood() {
-  let isHappy = true;
-
+let isHappy = false;
+function getMood(isHappy) {
   if (isHappy) {
     return "I am happy";
   } else {
     return "I am not happy";
   }
```
fowobi, line 18: Well done,Nice work

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,15 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+let userAge = 18;
```
munozirianni1988, line 7: Hi Christina, well done! :) these excersises were quite the challenge! Just a wee thing, I dont think is necessary here to give a value to userAge and isLoggedIn because the test will be doing that?

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,63 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+
+
+function printOddNumbers(limit) {
+  let lowNumber = 1;
+  while (lowNumber < limit) {
+    console.log(lowNumber);
+  lowNumber = lowNumber + 2;
+  }
+}
+
```
munozirianni1988, line 49: Very nice! I like your math :+1:

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,63 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
 
 /*
   Complete the buyTwoGetTheCheapestFree function: if user buys two items, the cheapest item will be free!
   The function should return the price to be paid once the discount is applied.
 */
-function buyTwoGetTheCheapestFree(price1, price2) {}
+
+function buyTwoGetTheCheapestFree(price1, price2) {
+  if (price1 && price2) {
```
munozirianni1988, line 57: Looks perfect! I think the code will run without line 57?

mandatory/2-function-creation.js
christina-mifsud: Hi Irianni! Thank you for your comment.
Yes you are correct! I repeated the mistake I made last week! I will keep it in mind in future!

mandatory/2-function-creation.js
christina-mifsud: Yes you are right thank you for pointing it out! 
In this case we do not need to check if the user is buying both items because it's implied in the price comparison (you cannot compare if you only have 1 item!)

mandatory/2-function-creation.js
```diff
@@ -93,15 +90,14 @@ function canRegister(age) {
...
 function countReverse(number) {
-  while (number > 0 && number <= 10) {
-    console.log(number);
-    number = number -1;
+  let count = number;
+  while (count > 0 && count <= number) {
+    console.log(count);
+    count = count - 1;
   }
 }
```
Muath-Alawadhi, line 100: Your attempt was great:  For loop in the function starts with i being equal to the input number, and decrements i by 1 at each iteration until i becomes 1. The console.log statement inside the loop logs the current value of i at each iteration, effectively producing a countdown of numbers from number down to 1.


function countReverse(number) {
  for (let i = number; i > 0; i--) {
    console.log(i);
  }
}

mandatory/1-fix-functions.js
Muath-Alawadhi, line 18: Defining the variable isHappy twice in the same code block, which is not allowed . Well, you need to remove the first definition of isHappy that is not used in the function.


function getMood(isHappy) {
  
  if (isHappy ) {
    return "I am happy";
  } else {
    return "I am not happy";
  }
}

mandatory/1-fix-functions.js
```diff
@@ -9,21 +9,18 @@
...
 
+let isBigEnough = 9;
 function greaterThan10(num) {
-  let isBigEnough;
-
-  if (isBigEnough) {
+  if (num > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
```
Muath-Alawadhi, line 26: The variable isBigEnough is defined twice variable inside the greaterThan10 function, which means that the function will always use the second variable and ignore the first one.

It should be : function greaterThan10(num) {
    let isBigEnough = num > 10;

  if (isBigEnough) {
    return "num is greater than 10";
  } else {
  }
}

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,15 @@
...
-function isAcceptableUser(userAge, isLoggedIn) {}
+let userAge = 18;
+let isLoggedIn = false;
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge >= 18 && isLoggedIn === true){
+return true;
+  } else {
+    return false;
+  }
+}
```
Muath-Alawadhi, line 15: Well done ^ ^ : The variables userAge and isLoggedIn outside of the isAcceptableUser function but not passing them as arguments when calling the function. 

  if (userAge >= 18 && isLoggedIn) {
    return true;
  } else {
    return false;
  }
}

Muath-Alawadhi (commented):
Well done * *

mandatory/1-fix-functions.js
```diff
@@ -9,21 +9,18 @@
   NOTE:Only make edits inside the function
 
 */
-
-function getMood() {
-  let isHappy = true;
-
+let isHappy = false;
```
jackgoldmclean, line 12: hey!
so what you've done here is define a variable called 'isHappy' and then below you've used it as a function parameter. This is called [Variable Shadowing](https://www.geeksforgeeks.org/variable-shadowing-in-javascript/) - it technically works but i can be confusing and unnecessary, because it means that the first use of `isHappy` doesn't mean anything. 
You don't need to define isHappy as a variable before the function - defining it as a parameter in the function means it is sort of like a placeholder for when you pass it in when calling the function later!

mandatory/1-fix-functions.js
```diff
@@ -9,21 +9,18 @@
...
+function getMood(isHappy) {
   if (isHappy) {
     return "I am happy";
   } else {
     return "I am not happy";
   }
 }
 
+let isBigEnough = 9;
```
jackgoldmclean, line 21: same here!

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,63 @@ function isAcceptableUser(userAge, isLoggedIn) {}
     is applieds and 142.5 should be returned)
 */
 
-function applyDiscount(totalPrice) {}
+function applyDiscount(totalPrice) {
+  let fullDiscount = (totalPrice * 10)/100;
+  let smallDiscount = (totalPrice * 5)/100;
+  if (totalPrice >= 200){
+  return newPrice = totalPrice - fullDiscount;
```
jackgoldmclean, line 30: looks good! 
the only small thing I would mention here (that won't really make much difference but could be useful to think about) is that smallDiscount is calculated/defined but never used unless the totalPrice is less than 200 - maybe it should go in the if blocks like 
```
if (totalPrice >= 200){
  let fullDiscount = (totalPrice * 10)/100;
  return newPrice = totalPrice - fullDiscount;
} else {
  let smallDiscount = (totalPrice * 5)/100;
  return newPrice = totalPrice - smallDiscount;
  }
}
  return newPrice = totalPrice - fullDiscount;
```

this just means we aren't calculating anything we don't need so it will be a bit more efficient!

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,63 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
-function printOddNumbers(limit) {}
+*/
+
+
+
+function printOddNumbers(limit) {
+  let lowNumber = 1;
+  while (lowNumber < limit) {
+    console.log(lowNumber);
+  lowNumber = lowNumber + 2;
```
jackgoldmclean, line 46: looks good! 
a shorter way of writing (but no different do what you did) is
`lowNumber += 2`

mandatory/2-function-creation.js
jackgoldmclean, line 57: good check!

mandatory/2-function-creation.js
```diff
@@ -45,7 +90,14 @@ function canRegister(age) {}
   )
 */
 
-function countReverse(number) {}
+function countReverse(number) {
+  let count = number;
+  while (count > 0 && count <= number) {
```
jackgoldmclean, line 95: do we need the `count <= number` check if count starts at number and is reduced?

jackgoldmclean (commented):
really good work 🔥

PR #416, opened 2023-02-25
LONDON_10 | SAIM KORKMAZ | JS1-2

nsaimk:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,8 +10,7 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
 
   if (isHappy) {
     return "I am happy";
@@ -21,9 +20,8 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
 
-  if (isBigEnough) {
+  if (num > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -1,160 +1,108 @@
 /*
-  Complete the function to return a boolean describing whether a user is acceptable. 
-  To be acceptable, two conditions must be fulfilled
-    1. the user should be 18 or older
-    2. the user must be logged in
-*/
-function isAcceptableUser(userAge, isLoggedIn) {}
-
-/*
-  Complete the function to apply discount percent based on how much is totalPrice in user cart.
-    - When the total price is greater than 200, a 10% discount should be applied
-    - When the total price is less than 200, a 5% discount should be applied
-
-  The function should return the new price to be paid (e.g. if the totalPrice is 150, a 5% discount
-    is applieds and 142.5 should be returned)
+Complete the function to check if the variable `num` satisfies the following requirements:
+- is a number
+- is even
+- is less than or equal to 100
+Tip: use logical operators
 */
 
-function applyDiscount(totalPrice) {}
+function validate(num) {
+  //The typeof operator returns a string indicating the type of the operand's value.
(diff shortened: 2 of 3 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,8 +10,7 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
 
   if (isHappy) {
```
berkeli, line 15: There's a pattern in programming to simplify `if .. else ` statements. You can rewrite this block as follows:

```
 if (isHappy) {
    return "I am happy";
  }

return "I am not happy";
```
the pattern is call early return statement, and you can read more here https://medium.com/swlh/return-early-pattern-3d18a41bba8

mandatory/2-function-creation.js
```diff
@@ -1,160 +1,108 @@
...
-*/
-
-function applyDiscount(totalPrice) {}
+function validate(num) {
+  //The typeof operator returns a string indicating the type of the operand's value.
+  if (typeof num == "number" && num % 2 == 0 && num <= 100) {
+    return true;
+  } else {
+    return false;
+  }
```
berkeli, line 15: Same goes here for return early statement

mandatory/2-function-creation.js
```diff
@@ -1,160 +1,108 @@
...
-  - if the person is older than 12 and younger than 90 it should return "You Can Register"
-  - if the person is 90 or older it should return "You Don't Need To Register"
-*/
-function canRegister(age) {}
+function formatPercentage(num) {
+  //num = Math.min(num, 100); // replace number with 100 if it's greater than 100
+  if (num >= 100) {
+    num = 100;
+  }
+  return num.toFixed(2).replace(/\.?0+$/, "") + "%";
```
berkeli, line 31: This seems overly complicated. 
You can rewrite this with `Math.round` which will not produce the trailing zeros you are removing. 

```
return `${Math.round(num * 100) / 100}%`
```
will produce the correct answer and you do not need to use regex. But if you want to use regex, it's always best to leave a comment explaining what it does becaue regex is not easy to understand

mandatory/2-function-creation.js
```diff
@@ -1,160 +1,108 @@
...
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
 
-/*
-  Complete the function to apply discount percent based on how much is totalPrice in user cart.
-    - When the total price is greater than 200, a 10% discount should be applied
-    - When the total price is less than 200, a 5% discount should be applied
-
-  The function should return the new price to be paid (e.g. if the totalPrice is 150, a 5% discount
-    is applieds and 142.5 should be returned)
```
berkeli, line 15: This exercise is missing?

mandatory/2-function-creation.js
```diff
@@ -1,160 +1,108 @@
...
-*/
-
-function applyDiscount(totalPrice) {}
```
berkeli, line 18: This exercise was removed?

mandatory/2-function-creation.js
```diff
@@ -1,160 +1,108 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
 
 /*
-  Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
-  */
-function printOddNumbers(limit) {}
```
berkeli, line 23: Same as before, this was removed?

berkeli (changes requested):
I left a few comments, but I think the 2-function-creation.js file has quite a few exercises removed and related tests removed as well. 

Is there a reason why these were removed?

PR #423, opened 2023-02-25
GLASGOW6_IRIANNI_MUNOZ-JAVASCRIPT-2

munozirianni1988:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-factorial.js
+++ b/extra/1-factorial.js
@@ -9,7 +9,13 @@
 */
 
 function factorial(input) {
-  // TODO
+  let result = input;
+  while (input > 1) {
+    input--; 
+    result *= input;  
+    
+  }
+  return result;
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,15 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge >= 18 && isLoggedIn == true) { 
+    return true;
+
+  }
+  else {
+    return false;
+  }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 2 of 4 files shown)


mandatory/2-function-creation.js
```diff
@@ -15,12 +23,36 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  let number = 1;
+  while (number != limit) {
+    if (number % 2 != 0) {
+      console.log (number);
+
+    }
+
+    number ++;
+  }
+}
```
Fatemeh-sadat-Torabi, line 55: the function you wrote is good and It works, but the function you wrote, in each run of the while loop, will check a condition " if " statement. 
For instance, if the "limit" is 100, the loop will run 100 times, and each time it will check the condition of " if ".  You can use another method that removes the if statement! 

```js
function printOddNumbers(limit) {
  let number = 1;
  while (number <= limit) {
    console.log(number);
    number += 2;
  }
}
```
this method is better because executing this will take less time! 
In fact, these two methods will be different when the " limit " is enormous, for example when you want to call printOddNumbers(1000000) and the while loop will run for 1000000 times. So imagine checking the " if " condition takes 1 second, in this case, the first method is 1 million seconds slower than the second method!

extra/1-factorial.js
```diff
@@ -9,7 +9,13 @@
...
 function factorial(input) {
-  // TODO
+  let result = input;
+  while (input > 1) {
+    input--; 
+    result *= input;  
+    
+  }
+  return result;
 }
```
Fatemeh-sadat-Torabi, line 19: You used an iterative approach, it is very good and it works and I liked it, but you can also do it like below: 
```js
function factorial(input) {
  if (input < 2) {
    return 1;
  } else {
    return input * factorial(input - 1);
  }
}
```
This approach is called the recursive approach! You will learn more about it in the future.

Fatemeh-sadat-Torabi (approved):
I reviewed the codes, well done!

gregdyke (changes requested):
Because we want to leave the main branch with exercises *not* solved, we don't typically approve any PRs, just leave comments (so i'm pressing "request changes" to prevent accidental merging)

mandatory/2-function-creation.js
munozirianni1988: Thank you Fatemeh! I understand your explanation very well. Thank you for reviewing my code :)

extra/1-factorial.js
munozirianni1988: Oh this looks very interesting! I will be reading about recursive approach later today after doing the homework. Thank you for pointing this out! 👍 :)

mandatory/2-function-creation.js
Fatemeh-sadat-Torabi: You're welcome!

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,68 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
 
 /*
   Complete the function to determine if it is suitable for a person to register based on their age!
   - if the person is 12 or younger it should return "You Are Too Young To Register"
   - if the person is older than 12 and younger than 90 it should return "You Can Register"
   - if the person is 90 or older it should return "You Don't Need To Register"
 */
-function canRegister(age) {}
+function canRegister(age) {
```
lazar-eva, line 76: Great job Irianni! Looks good to me ⭐️

PR #426, opened 2023-02-26
Glasgow Class 6 Melese Berehannu JavaScript-Core-1-Coursework-Week 2

Melesegithub:
Mandatory JS Week 2

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-factorial.js
+++ b/extra/1-factorial.js
@@ -9,7 +9,12 @@
 */
 
 function factorial(input) {
-  // TODO
+  // TODOlls
+  for (let i = input - 1; i > 0; i--) {
+      input = input * i;
+    
+  }
+  return input;
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,8 +10,9 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(bool) {
+  let isHappy = bool;
+
 
   if (isHappy) {
     return "I am happy";
@@ -19,11 +20,11 @@ function getMood() {
     return "I am not happy";
   }
 }
-
+//we need to call a function 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num;
 
-  if (isBigEnough) {
+  if (isBigEnough > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,13 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if(userAge >= 18 && isLoggedIn === true){
+    return true;
+  }else{
+    return false;
+  }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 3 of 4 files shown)


extra/1-factorial.js
```diff
@@ -9,7 +9,12 @@
 */
 
 function factorial(input) {
-  // TODO
+  // TODOlls
+  for (let i = input - 1; i > 0; i--) {
+      input = input * i;
+    
+  }
+  return input;
```
annacollins85, line 17: Great 👏

mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,21 @@
...
+function getMood(bool) {
+  let isHappy = bool;
+
 
   if (isHappy) {
     return "I am happy";
   } else {
     return "I am not happy";
   }
 }
```
annacollins85, line 22: This works fine, but we can make it even simpler. We can just pass the parameter directly into the if statements (we can call it what we like, here I have called it `isHappy` as before).
```
function getMood(isHappy) {
  if (isHappy) {
    return "I am happy";
  } else {
    return "I am not happy";
  }
}
```

mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,21 @@
...
-
+//we need to call a function 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num;
 
-  if (isBigEnough) {
+  if (isBigEnough > 10) {
```
annacollins85, line 27: `isBigEnough` implies either yes, it is, or no, it isn't, i.e. a boolean. So we should set this variable to 
```
const isBigEnough = num > 10
```
And then in the if statement we only need to pass `if (isBigEnough)`. You solution works but the variable name doesn't explain properly what the variable is representing since you just set it to number.

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,13 @@
...
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if(userAge >= 18 && isLoggedIn === true){
+    return true;
+  }else{
+    return false;
+  }
+}
```
annacollins85, line 14: Nice 👌

mandatory/2-function-creation.js
```diff
@@ -15,26 +21,69 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+function applyDiscount(totalPrice) {
+  if (totalPrice > 200){
+    newPrice = totalPrice - (0.1 * totalPrice);
+
+  }
+  else{
+    newPrice = totalPrice - (0.05 * totalPrice);
+  }
+ return newPrice;
+}
```
annacollins85, line 33: Great 👏

mandatory/2-function-creation.js
```diff
@@ -15,26 +21,69 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+    number = number + 1;
+    if (number === 1){
+    console.log(number);
+  }else if(number % 2 !== 0){
+    console.log(number)
+  }
+    
+
+  }
+}
```
annacollins85, line 50: Nice. This works but we can make it even simpler because `1%2` is equal to 1. So it will go into your if statement.
```
function printOddNumbers(limit) {
  number = 0;
  while(number < limit){
    number = number + 1;
    if(number % 2 !== 0) {
      console.log(number)
    }
  }
}

File: mandatory/2-function-creation.js [hunk 7]
```diff
@@ -15,26 +21,69 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  
+  }else if(price1 < price2){
+    return price2;
+
+  }
+  else{
+    return price1;
+  }
+
+}
```
annacollins85, line 68: As discussed in the buddy session. We can use something called a [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) here to make this super short.
```
function buyTwoGetTheCheapestFree(price1, price2) {
 return price1 < price2 ? price2 : price1
}
```
This works since if the prices are the same then it will just return one of them (doesn't matter which), in our case, `price1 < price2` will not be true if they are the same, so our ternary operator will return what's after the `:`, which is `price1`

mandatory/2-function-creation.js
```diff
@@ -15,26 +21,69 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  if (age <=12){
+    return "You Are Too Young To Register";
+  }
+  else if (age >= 90){
+    return "You Don't Need To Register";
+  }
+  else{
+    return "You Can Register";
+  }
+}
```
annacollins85, line 86: 👏

mandatory/2-function-creation.js
```diff
@@ -45,7 +94,14 @@ function canRegister(age) {}
...
 
-function countReverse(number) {}
+function countReverse(number) {
+  let count;
+ for (count = number; count > 0; count--) { 
+  
+   console.log(count) ;
+   
+   }
+}
```
annacollins85, line 104: 👏

annacollins85 (commented):
Great work! You have shown you understand conditionals well. Well done 🙂

PR #429, opened 2023-02-26
Glasgow Class 6 - nasir ali - JS1 - Week 2 

ali-nasir-ali:

# Your Details

- Your Name: nasir ali
- Your City: glasgow
- Your Slack Name: nasir ali

# Homework Details

- Module: javascript
- Week: 2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-factorial.js
+++ b/extra/1-factorial.js
@@ -10,6 +10,11 @@
 
 function factorial(input) {
   // TODO
+  var factored = 1;
+  for (var i = input; i > 0; i--) {
+    factored = factored * i;
+  }
+  return factored;
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -16,6 +16,7 @@ function getMood() {
   if (isHappy) {
     return "I am happy";
   } else {
+    //
     return "I am not happy";
   }
 }
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,21 +10,16 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function getMood(value) {
+  if (value === true) {
     return "I am happy";
   } else {
-    //
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
-
-  if (isBigEnough) {
+  if (num > 10) {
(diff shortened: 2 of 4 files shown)


mandatory/2-function-creation.js
```diff
@@ -15,26 +21,52 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+    return totalPrice - (totalPrice / 100) * 5;
+  }
+}
 
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
   */
-function printOddNumbers(limit) {}
+function printOddNumbers(limit) {
+  for (var i = 0; i <= limit; i++) {
```
munozirianni1988: Hi Nasir, well done! Great job, but I think we need to use a while loop for this exercise.

ali-nasir-ali: thank you Irianni ,i will fix that right now.

extra/1-factorial.js
```diff
@@ -10,6 +10,11 @@
 
 function factorial(input) {
   // TODO
```
Joemwa, line 12: This comment is not needed anymore.

Joemwa (commented):
comment

extra/1-factorial.js
```diff
@@ -10,6 +10,11 @@
...
+  var factored = 1;
```
lazar-eva, line 13: Hey Nasir, great job on getting all the exercises done! Just a little suggestion - maybe try to replace` var` with `const` and `let ` in the future? I think the new convention (since ES6) is to use them instead of` var `.  https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/

mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,16 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function getMood(value) {
+  if (value === true) {
```
hussein-alsayed, line 14: probably u know about this buddy, but I have to let you know so i can tick the commenting assignment box off the list :)

you could've just left value without if statement, because as long as the value is not one of the falsies it always elevates to true (or this is how i undrerstand it).

best, 
see u tomorrow.

mandatory/2-function-creation.js
Dedekind561: Also please use the `let` keyword for the `i` variable declaration

PR #431, opened 2023-02-27
Glasgow Class_6 - Rahma Berhan - JS1 - WEEK_2 

RahmaB1:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -9,13 +9,15 @@
   NOTE:Only make edits inside the function
 
 */
+let mood;
 
-function getMood() {
-  let isHappy = true;
+function getMood(mood) {
+  let isHappy;
 
-  if (isHappy) {
+  if (mood === true) {
     return "I am happy";
-  } else {
+  } 
+  else {
     return "I am not happy";
   }
 }
@@ -23,7 +25,7 @@ function getMood() {
 function greaterThan10(num) {
   let isBigEnough;
 
-  if (isBigEnough) {
+  if (num > 10 ) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -9,23 +9,17 @@
   NOTE:Only make edits inside the function
 
 */
-let mood;
 
 function getMood(mood) {
-  let isHappy;
-
   if (mood === true) {
     return "I am happy";
-  } 
-  else {
+  } else {
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
-
-  if (num > 10 ) {
+  if (num > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,19 +4,12 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) 
-{
-  if(userAge >= 18 && isLoggedIn===true)
-  {
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge >= 18 && isLoggedIn === true) {
     return true;
-  }
-  else
-  {
+  } else {
     return false;
   }
-
-
-
 }
 
 /*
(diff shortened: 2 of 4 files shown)


mandatory/1-fix-functions.js
```diff
@@ -9,21 +9,23 @@
   NOTE:Only make edits inside the function
 
 */
+let mood;
```
annacollins85: We don't need to define `mood` here. It's passed into the function below as `mood`, but we could call it anything we want. It's just a name we give to whatever we pass to the function when we call it.

mandatory/1-fix-functions.js
```diff
@@ -9,21 +9,23 @@
...
 
-function getMood() {
-  let isHappy = true;
+function getMood(mood) {
+  let isHappy;
```
annacollins85: We don't need this variable here since we are not using it. You have renamed the function parameter to `mood` which is totally fine and it replaces the need for the `isHappy` variable.

mandatory/1-fix-functions.js
```diff
@@ -9,21 +9,23 @@
...
+  else {
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
   let isBigEnough;
 
-  if (isBigEnough) {
+  if (num > 10 ) {
```
annacollins85: Here you can remove the `let isBigEnough` since we are not using it. But another option would be to say
```
let isBigEnough = num > 10
if (isBigEnough){
...
```

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,20 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) 
+{
+  if(userAge >= 18 && isLoggedIn===true)
```
annacollins85: Great!

mandatory/2-function-creation.js
```diff
@@ -15,26 +28,87 @@ function isAcceptableUser(userAge, isLoggedIn) {}
     is applieds and 142.5 should be returned)
 */
 
-function applyDiscount(totalPrice) {}
+function applyDiscount(totalPrice)
```
annacollins85: 👏

mandatory/2-function-creation.js
```diff
@@ -15,26 +28,87 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+}
+
+
 
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
   */
-function printOddNumbers(limit) {}
+function printOddNumbers(limit) 
+{
```
annacollins85: Just a comment on coding style. Generally we would keep this opening `{` on the line above after the function parameters, e.g.
`function printOddNumbers(limit) {`
You can install Prettier on VSCode and set up format on save so that you always have your code in a consistent style inline with conventions. I can help with this if you need help. Here is an article explaining how https://www.educative.io/answers/how-to-set-up-prettier-and-automatic-formatting-on-vs-code

annacollins85 (commented):
Nice work! You have clearly got the hang of conditional statements 👏 

The only thing I would say to improve is to make sure your code is formatted properly. Installing the Prettier VS Code extension will do this for you. Let me know if you need some help.

mandatory/1-fix-functions.js
RahmaB1: Oh I see. Thanks :)

mandatory/1-fix-functions.js
RahmaB1: You are right. I don't even remember why I had it lols.

mandatory/1-fix-functions.js
RahmaB1: so, this way we can save the answer from the comparison ( num > 10 ) into isBigEnough which is going to be true or false, is this correct? 

Thanks

mandatory/2-function-creation.js
RahmaB1: I thought it would be easier to read this way. Definitely will use Prettier to let my code be formatted as standard and to look much prettier ; ) Thanks

PR #433, opened 2023-02-27
London_10-Saliha_Popal-JavaScript-Core-1-Coursework-Week2

SalihaPopal:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-factorial.js
+++ b/extra/1-factorial.js
@@ -9,7 +9,11 @@
 */
 
 function factorial(input) {
-  // TODO
+  let number = 1;
+  for(let i = 1; 1<= number; i++){
+    number*=1
+  }
+  return number;
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,9 +10,7 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
-
+function getMood(isHappy) {
   if (isHappy) {
     return "I am happy";
   } else {
@@ -21,9 +19,9 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = 10;
 
-  if (isBigEnough) {
+  if (isBigEnough < num) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,13 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if(userAge >=18 && isLoggedIn){
+    return true
+  } else{
+    return false
+  }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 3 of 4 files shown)


SalihaPopal:
Hello there, could you please help me why the PR has red x and it has not been merged yet? Although all tests have been passed. How can I fix it?
Thank you.

mandatory/1-fix-functions.js
```diff
@@ -21,9 +19,9 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = 10;
```
SalihaPopal, line 22: The value of isBigEnough is declared = num and greater then 10.
let isBigEnough = num > 10

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,13 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if(userAge >=18 && isLoggedIn){
+    return true
```
SalihaPopal, line 9: let isAcceptable = userAge >=18 && isLoggedIn; inside of the isAcceptableUser function. No need for if and else function.
return isAccptable; 
Here should be just return isAccptable;

SalihaPopal:
Hello there, this PR has the same problem. Could you please help me to fix it?
Thank you.

extra/1-factorial.js
```diff
@@ -9,7 +9,11 @@
 */
```
moneyinthesky, line 10: Well done for attempting this - it's a tough one!
A couple of hints:
1. Are you using the `input` parameter in your function?
2. Try stepping through your code with the VsCode debugger, and see if it's working as you expect it to. (let us know if you need help with that 😄 )

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,13 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
```
moneyinthesky, line 6: This looks good 👍 
Can you think of how to make this code even shorter? Do you need the `if` statement? Can you do it with only a `return` statement?

mandatory/2-function-creation.js
```diff
@@ -15,26 +21,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
     is applieds and 142.5 should be returned)
 */
```
moneyinthesky, line 23: Nice! When writing a function, it's important to think about what values the parameter could be. In this case, what happens when `totalPrice` is equal to 200? What will your function return?

mandatory/2-function-creation.js
```diff
@@ -15,26 +21,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  }
+  if(totalPrice < 200){
+    totalPrice -= totalPrice * 0.5;
+    return totalPrice;
+  }
+}
 
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
   */
```
moneyinthesky, line 37: This one is almost there - but there's a couple of small problems.
- In the code below you're doing `number % 2 == 1;` - did you want to use an `if` statement here?
- Right now, this code will result in an infinite loop (this means the while loop will never end). I would suggest you use the VsCode debugger to step through the code line by line. Check what's happening to the `number` variable while you are doing this. (let us know if you need help with that 😄)

mandatory/2-function-creation.js
```diff
@@ -15,26 +21,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+      console.log(number);
+    }
+    number++
+  }
+
 
 /*
   Complete the buyTwoGetTheCheapestFree function: if user buys two items, the cheapest item will be free!
   The function should return the price to be paid once the discount is applied.
 */
```
moneyinthesky, line 51: Nice 👍

mandatory/2-function-creation.js
```diff
@@ -15,26 +21,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+    return price1;
+  }
+}
 
 /*
   Complete the function to determine if it is suitable for a person to register based on their age!
   - if the person is 12 or younger it should return "You Are Too Young To Register"
   - if the person is older than 12 and younger than 90 it should return "You Can Register"
   - if the person is 90 or older it should return "You Don't Need To Register"
 */
```
moneyinthesky, line 65: Looks perfect!

mandatory/2-function-creation.js
```diff
@@ -45,7 +83,13 @@ function canRegister(age) {}
   )
 */
```
moneyinthesky, line 85: Interesting implementation 😄 
Here you are calling the function from inside itself - this is called a recursive function.
It works! But for extra practice, can think of how you would do this with a loop instead? Maybe a `while` or `for` loop?

moneyinthesky:
Good effort on this coursework @SalihaPopal!
I have left a few comments for you 😄

PR #440, opened 2023-03-01
Glasgow_6-Muath- Alawadhi-JavaScript-Core-1-Coursework-Week2

Muath-Alawadhi:

# Your Details

- Your Name: Muath Alawadhi
- Your City: Glasgow
- Your Slack Name: Muath Alawadhi

# Homework Details

- Module:JavaScript-Core-1-Coursework
- Week: 2

# Notes

- What did you find easy?

- What did you find hard?
- It was a bit difficult and took a lot of time
- 

- Any other notes?

### DIFF
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,14 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge >= 18 && isLoggedIn) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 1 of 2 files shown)


mandatory/2-function-creation.js
```diff
@@ -15,26 +22,69 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  }
+}
 
 /*
   Complete the buyTwoGetTheCheapestFree function: if user buys two items, the cheapest item will be free!
   The function should return the price to be paid once the discount is applied.
 */
-function buyTwoGetTheCheapestFree(price1, price2) {}
+function buyTwoGetTheCheapestFree(price1, price2) {
+  if (price1 <= 0 || price2 <= 0) {
```
gregdyke, line 54: A function that is supposed to return a number should never return a string if there is a problem - that just kicks the problem down the road.

We will see later how to handle inputs that have a problem - for now, just assume that inputs are reasonable unless explicitly specified otherwise

mandatory/2-function-creation.js
```diff
@@ -15,26 +22,69 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+    return "Invalid prices";
+  }
+
+  let totalPrice = price1 + price2;
```
gregdyke, line 58: This is a very complicated way of doing this. Could you think of a simpler way that doesn't involve adding an subtracting?

mandatory/2-function-creation.js
```diff
@@ -45,7 +95,11 @@ function canRegister(age) {}
   )
 */
 
-function countReverse(number) {}
+function countReverse(number) {
+  for (let i = number; i > 0; i--) {
```
gregdyke, line 99: nice use of a for loop!

gregdyke: can you now re-write it with a while loop? (hint: for loops and while loops are exactly equivalent to each other: all for loops can be written as while loops and all while loops can be written as for loops)

gregdyke (commented):
Good job Muath! You seem to not have format on save working correctly for you

mandatory/2-function-creation.js
```diff
@@ -15,26 +22,69 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
 
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
   */
-function printOddNumbers(limit) {}
+function printOddNumbers(limit) {
+  let num = 1;
+  
+  while (num <= limit) {
+    if (num % 2 !== 0) {
```
christina-mifsud, line 42: Great solution! An alternative would be to increment the num by 2 (if you start counting at 1 and add 2, the result would always be an add number - 1+2 = 3, 3+2 = 5 etc.) and you would be able to eliminate the if statement.

christina-mifsud (commented):
Review completed as per required coursework. Well done!

mandatory/2-function-creation.js
Muath-Alawadhi: OK thanks

mandatory/2-function-creation.js
Muath-Alawadhi: maybe like this 
function countReverse(number) {
  let i = number;
  while (i > 0) {
    console.log(i);
    i--;
  }
}

mandatory/2-function-creation.js
Muath-Alawadhi: that's a good point! thanks

mandatory/2-function-creation.js
Muath-Alawadhi: I just realized it was really complicated.  

function buyTwoGetTheCheapestFree(price1, price2) {
  if (price1 > price2) {
    return price1;
  } else {
    return price2;
  }
}

PR #442, opened 2023-03-01
Glasgow_6 - HERISH  TURKI - JS -Core-1-Coursework-Week2 

HereshT:
Glasgow_6 - HERISH  TURKI - JS -Core-1-Coursework-Week2

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -13,7 +13,7 @@
 function getMood(isHappy) {
   let isHappy = true;
 
-  if (isHappy) {
+  if (true) {
     return "I am happy";
   } else {
     return "I am not happy";
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -18,7 +18,7 @@ function getMood(isHappy) {
   }
 }
 
-////////////////////
+//////////s/////////////////
 
 function greaterThan10(num) {
   if (num > 10) {
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,20 +10,18 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function getMood(isHappy) {
+  if (isHappy === true) {
     return "I am happy";
   } else {
     return "I am not happy";
   }
 }
 
-function greaterThan10(num) {
-  let isBigEnough;
+////////////////////
 
-  if (isBigEnough) {
+function greaterThan10(num) {
+  if (num > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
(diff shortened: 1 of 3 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,18 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function getMood(isHappy) {
+  if (isHappy === true) {
```
SallyMcGrath, line 14: This works and is a good answer. 

It might be interesting now to explore ways to refactor and simply this function. 

`if (isHappy === true)`

If isHappy is the same as true, do you need to make this check? What does isHappy _evaluate_ to? (Replace the left hand side with the value of isHappy and read the code again. )

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,9 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
```
SallyMcGrath, line 7: Another great answer.

How might you refactor this? Try writing this function a few different ways and thinking about it (this is also fun to do in Codewars).

mandatory/2-function-creation.js
```diff
@@ -15,26 +17,67 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+
+// or
+
+// function applyDiscount(totalPrice) {
+//   if (totalPrice > 200){
+//     return totalPrice * 0.9;
+//   } else {
+//     return totalPrice * 0.95;
+//   }
+// }
```
SallyMcGrath, line 38: Fab. I definitely prefer the code you chose over the commented out solution. Can you guess why?

mandatory/2-function-creation.js
```diff
@@ -45,7 +88,11 @@ function canRegister(age) {}
   )
 */
 
-function countReverse(number) {}
+function countReverse(number) {
```
SallyMcGrath, line 91: I like this decrementing solution. I think the example solution uses a while loop, but I prefer yours! ;)

SallyMcGrath (commented):
Lovely work @HereshT . Thanks for your hard work and effort here. I've dropped a few comments and asked a few questions. Take a look and have a think.

mandatory/2-function-creation.js
HereshT: Regarding my approach to these tasks, I set aside my coding skills and knowledge and write the possible logical solutions using sudo language. Then, I use code to implement them. 

Back to your - why question -  I feel that the first solution is the most suitable since it is easy to understand, clear, and allows for the implementation of different discounts if needed.

PR #446, opened 2023-03-01
Glasgow 6 - Hussein Al-Sayed - JS - week 2

hussein-alsayed:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,7 +10,7 @@
 
 */
 
-function getMood() {
+function getMood(isHappy) {
   let isHappy = true;
 
   if (isHappy) {
@@ -21,7 +21,7 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = Number > 10;
 
   if (isBigEnough) {
     return "num is greater than 10";
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -11,7 +11,6 @@
 */
 
 function getMood(isHappy) {
-  let isHappy = true;
 
   if (isHappy) {
     return "I am happy";
@@ -21,7 +20,7 @@ function getMood(isHappy) {
 }
 
 function greaterThan10(num) {
-  let isBigEnough = Number > 10;
+  let isBigEnough = num > 10;
 
   if (isBigEnough) {
     return "num is greater than 10";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,9 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  return userAge >= 18 && isLoggedIn;
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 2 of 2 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,7 +10,7 @@
 
 */
 
-function getMood() {
+function getMood(isHappy) {
```
sztupy, line 13: while you realized that you need to add the isHappy as a parameter there is still the case of variable shadowing that blocks this function to behave properly

mandatory/1-fix-functions.js
```diff
@@ -21,7 +21,7 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = Number > 10;
```
sztupy: the idea is good, however care must be taken to make sure variable names that mean the same value are named exactly the same throughout

mandatory/2-function-creation.js
```diff
@@ -15,26 +17,56 @@ function isAcceptableUser(userAge, isLoggedIn) {}
     is applieds and 142.5 should be returned)
 */
 
-function applyDiscount(totalPrice) {}
+function applyDiscount(totalPrice) {
+  if (totalPrice > 200) {
+return totalPrice * 0.9;
```
sztupy: having proper indenting inside functions and blocks is usually a good way to make sure code is readable. You should make sure to use an editor that has auto-indenting enabled as that helps making sure everything is in order

mandatory/2-function-creation.js
```diff
@@ -15,26 +17,56 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
   The function should return the price to be paid once the discount is applied.
 */
-function buyTwoGetTheCheapestFree(price1, price2) {}
+function buyTwoGetTheCheapestFree(price1, price2) {
+  if (price1 > price2) {
+return price1;
+} else if (price2 > price1) {
+return price2;
+} else {
+  return price1||price2;
```
sztupy: If the price is the same then you can simply return whicever value

mandatory/2-function-creation.js
```diff
@@ -15,26 +17,56 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
 
 /*
   Complete the function to determine if it is suitable for a person to register based on their age!
   - if the person is 12 or younger it should return "You Are Too Young To Register"
   - if the person is older than 12 and younger than 90 it should return "You Can Register"
   - if the person is 90 or older it should return "You Don't Need To Register"
 */
-function canRegister(age) {}
+function canRegister(age) {
+  if (age < 13) {
```
sztupy: Prefer using `<=` and `>=` in case the text is defined that way to avoid differences between the specification and the solution

mandatory/2-function-creation.js
hussein-alsayed: true! I thought this way i'd be showing that i'm aware there's no difference. but i guess code length  and simplicity matters more.. edited now  👍

mandatory/2-function-creation.js
hussein-alsayed: good point, thank you

mandatory/2-function-creation.js
hussein-alsayed: i'm using vsc, is there any extensions i can use for this?

mandatory/1-fix-functions.js
hussein-alsayed: true, fixed now  👍

mandatory/1-fix-functions.js
hussein-alsayed: still trying to understand these differences and their behaviour. however i believe i fixed it now  👍

PR #450, opened 2023-03-04
glasgow class 6-siver omar-javascript-week2

siveromar:

# Your Details

- Your Name: Siver
- Your City: Glasgow
- Your Slack Name: siver omar

# Homework Details

- Module: JS
- Week: 2

### DIFF
--- a/extra/1-factorial.js
+++ b/extra/1-factorial.js
@@ -10,6 +10,13 @@
 
 function factorial(input) {
   // TODO
+  if (input < 0) 
+  return -1;
+else if (input == 0) 
+return 1;
+else {
+return (input * factorial(input - 1));
+}
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,8 +10,8 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
+
 
   if (isHappy) {
     return "I am happy";
@@ -23,13 +23,13 @@ function getMood() {
 function greaterThan10(num) {
   let isBigEnough;
 
-  if (isBigEnough) {
+  if (num > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
   }
 }
-
+// npm test -- --testPathPattern 1-fix-function
 /* ======= TESTS - DO NOT MODIFY ===== */
 
 describe("getMood", () => {
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,15 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+// npm test -- --testPathPattern 2-function-creation
+function isAcceptableUser(userAge, isLoggedIn) {
+  if( userAge >= 18 && isLoggedIn){
+    return true
+  }
+  else{
+    return false
+  }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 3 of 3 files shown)


mandatory/1-fix-functions.js
```diff
@@ -23,13 +23,13 @@ function getMood() {
 function greaterThan10(num) {
   let isBigEnough;
```
msimmdev, line 24: Nitpick: What you've done is correct, but it leaves a pointless variable isBigEnough that is never used. One possibility could be to use the isBigEnough variable to store the condition output. e.g.
```
let isBigEnough = num > 10;

if (isBigEnough) {
...
```
Or alternatively you could have just deleted the `isBigEnough` variable.

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,15 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+// npm test -- --testPathPattern 2-function-creation
+function isAcceptableUser(userAge, isLoggedIn) {
+  if( userAge >= 18 && isLoggedIn){
```
msimmdev, line 9: Nitpick: While this is OK, it's slightly better to do isLoggedIn === true as this type checks isLoggedIn to make sure it's passed into the function as a boolean.

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,56 @@ function isAcceptableUser(userAge, isLoggedIn) {}
     is applieds and 142.5 should be returned)
 */
 
-function applyDiscount(totalPrice) {}
+function applyDiscount(totalPrice,discount) {
```
msimmdev, line 26: Issue: There's no need to take `discount` as a parameter to this function. You should only define the inputs that are needed by the function or expected by the tests.

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,56 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  if (totalPrice >200) {
+    total= totalPrice-(totalPrice*0.1);
```
msimmdev, line 28: Nitpick: You've created a new variable `total`, this is OK but you should always declare a new variable with `let` (or `const` if it's appropriate)

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,56 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
+  }
+}
 
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
   */
-function printOddNumbers(limit) {}
+function printOddNumbers(limit) {
+  let count = 1;
+while(count<=limit){
```
msimmdev, line 41: Nitpick: The logic is correct, but the indentation seems incorrect here.

mandatory/2-function-creation.js
```diff
@@ -45,7 +83,13 @@ function canRegister(age) {}
   )
 */
 
-function countReverse(number) {}
+function countReverse(number) {
+  while( number >= 1){
```
msimmdev, line 87: Suggestion: create a new variable that you can loop over and decrement rather than using the input `number`. e.g.
```
let currentNumber = number;
while (currentNumber  >= minNum) {
  console.log(number);
  currentNumber--;
}
```
The reason for this is you still have access to the unmodified user input in case you need to use it in  a future modification of the function.
Also you'll find in the future lessons that for certain types passed to a function modifying the input data could have an impact on other parts of the program (this isn't a problem for integers though so wouldn't be an issue here)

extra/1-factorial.js
```diff
@@ -10,6 +10,13 @@
 
 function factorial(input) {
   // TODO
+  if (input < 0) 
+  return -1;
+else if (input == 0) 
+return 1;
+else {
+return (input * factorial(input - 1));
```
msimmdev, line 18: Issue: This works, but isn't implemented using a loop as requested in the description. This uses a recursive function, which is an advanced topic we haven't covered yet.

msimmdev (commented):
Well done on completing the mandatory exercises, it looks like you're getting to grips with the basic logic.

I've spotted a couple of issues, I'd suggest you take a look through those and if you've got time try and resolve the noted problems. There are a few other notes I've made where your code could be formatted a little better or suggestions of different ways to approach the problems. Take a read through and bear those in mind for future exercises.

PR #451, opened 2023-03-04
final change- week2

AkramIzedi:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,20 +10,17 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function getMood(happy) {
+  if (happy == true) {
     return "I am happy";
-  } else {
+  } else if (happy == false){
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
 
-  if (isBigEnough) {
+  if (num>10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,15 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge>=18 && isLoggedIn===true){
+    return true;
+  }else if (userAge>18 && isLoggedIn===true){
+  return false;
+}else {
+  return false;
+  }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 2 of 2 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,17 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
-
-  if (isHappy) {
+function getMood(happy) {
+  if (happy == true) {
```
selinfildis, line 14: Since booleans can only be true or false you shouldn't need to check if happy is equal to true or false. Checking the variable happy should be enough to determine whether the mood is happy or not.

You can read more on booleans here: https://www.w3schools.com/js/js_booleans.asp

mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,17 @@
...
     return "I am happy";
-  } else {
+  } else if (happy == false){
```
selinfildis, line 16: since happy can only be either true or false and we want to make sure to enter this block and return "I am not happy" when happy is false, the if else statement should be enough to determine what to return and we shouldn't need a `happy == false` check

mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,17 @@
...
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
 
-  if (isBigEnough) {
+  if (num>10) {
```
selinfildis, line 23: Nice one!

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,15 @@
...
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge>=18 && isLoggedIn===true){
+    return true;
+  }else if (userAge>18 && isLoggedIn===true){
+  return false;
+}else {
+  return false;
+  }
+}
```
selinfildis, line 15: The condition to accept the user or not is determining whether userAge >= 18 and if the user is logged in

`userAge >= 18`  -> this statement should return a boolean, so if you assign it to a variable, that variable's type should be a boolean. You can see the example code to see this here:

```js

let isUserAgeBigger = userAge >= 18;
console.log(typeof isUserAgeBigger); // this should be a boolean
console.log(isUserAgeBigger) // this should be the value of it
```

the variable isLoggedIn is already of type boolean, as you can see in the examples from the tests [here](https://github.com/CodeYourFuture/JavaScript-Core-1-Coursework-Week2/pull/451/files?diff=split&w=0#diff-6b32acd5b3a1f33e1f035a68477c6641df187cb5ee57e41615bb942d518ade86L52)


Since you only want a boolean return from this function, you could just evaluate the boolean expression, by returning the statement inside the first `if` condition, since it will only return true if the statement inside the if condition (`userAge>=18 && isLoggedIn===true`) is true. 

So you could just do: 
```js
    return userAge>=18 && isLoggedIn===true
}
```

and it should return the behaviour above is doing. 

Also since `isLoggedIn` is a boolean value, you do not need to check if it's equal to true or not. you should be able to directly use it in the conditions.

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
 */
 
-function applyDiscount(totalPrice) {}
+function applyDiscount(totalPrice) {
+  if (totalPrice>200){
+  return totalPrice *0.9;
+}else if (totalPrice<200){
+  return totalPrice * 0.95;
+ }
+}
```
selinfildis, line 32: Nice one! ☝️ 

This is exactly how the `else if` statement should be used, because we have a third condition where the `totalPrice` could be exactly equal to 200, but we haven't been requested to make a discount when the totalPrice is 200!

We should also `return totalPrice` just to ensure we have all cases covered (even though i'm aware there are no tests for this 😅  feel free to add one and try it out ^^)

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
 /*
   Complete the function to print to the console the odd numbers between 1 and limit (use a while loop):
   */
-function printOddNumbers(limit) {}
+function printOddNumbers(limit) {
+  for (let i = 1; i <= limit; i++) {
+    if (i % 2 === 1) {
+      console.log(i);
+    }
+  }
```
selinfildis, line 42: Awesome 👌

mandatory/2-function-creation.js
```diff
@@ -15,26 +23,58 @@ function isAcceptableUser(userAge, isLoggedIn) {}
...
 
 /*
   Complete the buyTwoGetTheCheapestFree function: if user buys two items, the cheapest item will be free!
   The function should return the price to be paid once the discount is applied.
 */
-function buyTwoGetTheCheapestFree(price1, price2) {}
+function buyTwoGetTheCheapestFree(price1, price2) {
+    if (price1 <= 0 || price2 <= 0) {
+      return "Invalid prices. Prices must be more than 0.";
+    }
```
selinfildis, line 53: Nice 😊

selinfildis (commented):
Hey @AkramIzedi - Could you also update your Pull Request title to fit the format please?

I've left a couple of comments related to the code as well ^^, 
overall, you're getting there!

PR #454, opened 2023-03-10
London 9 -Tuan Nguyen-JS-core1-week2

Jacknguyen4438:
# Your Details

- Your Name: Tuan
- Your City: London 9
- Your Slack Name: Tuan Nguyen

# Homework Details

- Module: JS-core 1
- Week: 2

# Notes

- What did you find easy? Understand of to use compare and logical operator

- What did you find hard? Doing with while loop

- What do you still not understand? No answer

- Any other notes? No answer

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,10 +10,10 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
+  //let isHappy = true;
 
-  if (isHappy) {
+  if (isHappy === true) {
     return "I am happy";
   } else {
     return "I am not happy";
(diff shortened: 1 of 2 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,20 +10,20 @@
...
+  if (isHappy === true) {
     return "I am happy";
   } else {
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough=num;
```
VicJerUk, line 24: Looks good but I'm wondering whether you can make it better? You did similar changes already in the `getMood` function. Would that work here too?

mandatory/2-function-creation.js
```diff
@@ -45,7 +80,12 @@ function canRegister(age) {}
   )
 */
 
-function countReverse(number) {}
+function countReverse(number) {
+  while (number >0){
+    console.log(number);
+  };
+  number=number-1;
+};
```
VicJerUk, line 88: Good effort but are you sure this is correct? When you ran this code locally did it work for you correctly? 
Look closely what you are trying to do with `number` parameter. Current code will result in never ending loop and will never complete is there a better way? 


Hint: Have a look at your `printOddNumbers` function.

VicJerUk (changes requested):

PR #456, opened 2023-03-25
London10-Onur-Atas-js1-week1

onurat:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,10 +10,10 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(mood) {
+  let isHappy = mood;
 
-  if (isHappy) {
+  if (isHappy === true) {
     return "I am happy";
   } else {
     return "I am not happy";
@@ -21,12 +21,13 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num;
 
-  if (isBigEnough) {
+  if (isBigEnough > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
+
   }
 }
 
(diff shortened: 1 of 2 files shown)


mandatory/1-fix-functions.js
```diff
@@ -10,23 +10,24 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(mood) {
+  let isHappy = mood;
```
VicJerUk, line 14: is this variable necessary? Could we compare against `mood` itself?

mandatory/1-fix-functions.js
```diff
@@ -10,23 +10,24 @@
...
+  if (isHappy === true) {
     return "I am happy";
   } else {
     return "I am not happy";
   }
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
+  let isBigEnough = num;
```
VicJerUk, line 24: Is this variable necessary? Could we compare against `num` itself?

mandatory/2-function-creation.js
```diff
@@ -4,7 +4,15 @@
...
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+function isAcceptableUser(userAge, isLoggedIn) {
+  if (userAge >= 18 && isLoggedIn === true) {
+    return true;
+  }
+  else {
+    return false;
+  }
```
VicJerUk, line 13: Great work so far, 

What you have written is correct. 

This is just a note, something to have fun with. There's another way you can achieve the same result. For example you could return the whole if statement condition and the output will be the same.


```
  return userAge >= 18 && isLoggedIn === true
}

VicJerUk (changes requested):

PR #457, opened 2023-03-30
London10-StellaDelMar_RodriguezFernandez-JavaScript-Core-1-Coursework-Week2

Stelladelmar:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -4,7 +4,15 @@
     1. the user should be 18 or older
     2. the user must be logged in
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+let userAge = 19;
+let isLoggedIn = true;
+function isAcceptableUser(userAge, isLoggedIn){
+  if (userAge >= 18 && isLoggedIn === true){
+  return true;
+ } else {
+  return false;
+ }
+}
 
 /*
   Complete the function to apply discount percent based on how much is totalPrice in user cart.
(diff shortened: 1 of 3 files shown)


mandatory/2-function-creation.js
```diff
@@ -4,48 +4,103 @@
...
     1
   )
 */
-
-function countReverse(number) {}
+//while loop
+//from 7 to 1
+let number = 7;
+function countReverse(number) {
+  while (number >= 10 ){
```
kwebster77, line 98: So for this one, using a while loop where will probably result in an infinite loop ([more on this here](https://flexiple.com/javascript/infinite-loops-javascript/)), if the number is greater than 10. Otherwise, this will not run as the number would be less than 10 (for example in your number = 7 example, this wouldn't return anything because the number is less than 10). 

You might want to think about using a for loop here: 

`for (let i = number; i > 0; i--)`

mandatory/2-function-creation.js
```diff
@@ -4,48 +4,103 @@
...
+
+function canRegister(age) {
+  if (age <= 12 ){
+    return "You Are Too Young To Register";
+  } else if (age < 90 && age > 12 ) {
+    return "You Can Register"; }
+    else if (age >= 90 ){
+    return "You Don't Need To Register"; }
+
+  }
```
kwebster77, line 80: Awesome! nice work on this one 🎉

mandatory/2-function-creation.js
```diff
@@ -4,48 +4,103 @@
...
-function buyTwoGetTheCheapestFree(price1, price2) {}
+let price1 = 700;
+let price2 = 500;
+function buyTwoGetTheCheapestFree(price1, price2) {
+  if(price1 > price2){
+    return price1;
+  } else if (price2 > price1){
+    return price2;
+  }
+  //console.log("price is");
```
kwebster77, line 57: nice job using an if statement here. 

LGTM! ([here](https://www.unosquare.com/blog/what-does-lgtm-mean-in-github-code-reviews/#:~:text=The%20letters%20LGTM%20stand%20for,use%20it%20after%20reviewing%20code.) is the meaning if this acronym!)

mandatory/2-function-creation.js
```diff
@@ -4,48 +4,103 @@
...
   */
-function printOddNumbers(limit) {}
-
+//function printOddNumbers(limit) {}
+let limit = 0;
+function printOddNumbers(limit) {
+while (limit <= 10){
+  console.log(limit);
+} limit = limit + 1;
+}
```
kwebster77, line 44: Looks like we might have another infinite loop here ([docs here](https://flexiple.com/javascript/infinite-loops-javascript/)). Maybe have a quick review on how while loops work? they can be a bit tricky.

mandatory/2-function-creation.js
```diff
@@ -4,48 +4,103 @@
...
-    is applieds and 142.5 should be returned)
+    is applied and 142.5 should be returned)
 */
-
-function applyDiscount(totalPrice) {}
-
+//let totalPrice = 150;
+function applyDiscount(totalPrice) {
+  if (totalPrice >200){
+    return (totalPrice - totalPrice * 0.01);
```
kwebster77, line 28: nice job with the if statement! 

Just might want to check your maths on this one 😉
0.01 would be 1% and 0.1 would be a 10%, but where is our 5% discount? 😜

mandatory/2-function-creation.js
```diff
@@ -4,48 +4,103 @@
...
 */
-function isAcceptableUser(userAge, isLoggedIn) {}
+let userAge = 19;
+let isLoggedIn = true;
+function isAcceptableUser(userAge, isLoggedIn){
+  if (userAge >= 18 && isLoggedIn === true){
+  return true;
+ } else {
+  return false;
+ }
```
kwebster77, line 14: Nice one!

kwebster77:
Hi @Stelladelmar! Thanks for putting in this PR! That's great :) Very happy you were able to get this in and a lot of really good work! I can see you have really been practicing. I have left some comments for you. The one thing I would say you might want to take another look at is your while loops. We can review this concept if you have any questions, but in the meantime take a look back through the videos and some of the documentation. 

Again, great work and keep it up! Looking forward to your next one :)

mandatory/2-function-creation.js
Stelladelmar: Thanks. I understand 😅

PR #459, opened 2023-04-05
London Class 10 - Damilola Odumosu - JavaScript-Core-1-Coursework-Week2

d-odumosu:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-fix-functions.js
+++ b/mandatory/1-fix-functions.js
@@ -10,8 +10,7 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
 
   if (isHappy) {
     return "I am happy";
@@ -21,9 +20,8 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
 
-  if (isBigEnough) {
+  if (num > 10) {
     return "num is greater than 10";
   } else {
     return "num is not big enough";
--- a/mandatory/2-function-creation.js
+++ b/mandatory/2-function-creation.js
@@ -26,7 +26,7 @@ return true;
 */
 
 function applyDiscount(totalPrice) {
-if(totalPrice >= 200){
+if(totalPrice > 200){
   return totalPrice - (totalPrice * 0.1)
 }
 else{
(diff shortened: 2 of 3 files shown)


mandatory/2-function-creation.js
```diff
@@ -4,37 +4,81 @@
...
   */
-function printOddNumbers(limit) {}
+function printOddNumbers(limit) {
+  let num = 1;
+  while (num < limit){
+
+ console.log(limit % 2 == 1 )
+ 
+  }
+}
```
VicJerUk, line 50: Good work so far, there's something you have missed here though.

This function will never complete. Have a look at your `countReverse` method you are missing a crucial piece of code needed for it to complete otherwise the `num` value will always stay at `1`.

VicJerUk (changes requested):

mandatory/1-fix-functions.js
```diff
@@ -10,8 +10,7 @@
 
 */
 
-function getMood() {
-  let isHappy = true;
+function getMood(isHappy) {
 
   if (isHappy) {
```
VicJerUk, line 15: Note: In an "early return" pattern you don't need to wrap your second return in an `else` condition.

so you could write your function in following fashion
```
  if (isHappy) {
    return "I am happy";
  } 
  return "I am not happy";

File: mandatory/1-fix-functions.js [hunk 3]
```diff
@@ -21,9 +20,8 @@ function getMood() {
 }
 
 function greaterThan10(num) {
-  let isBigEnough;
 
-  if (isBigEnough) {
+  if (num > 10) {
```
VicJerUk, line 24: You also can apply above pattern here taking there are only 2 outcomes of your function.

mandatory/1-fix-functions.js
VicJerUk: Again this is only a suggestion what you have written is correct.

VicJerUk (changes requested):
There are some issues in your code preventing it from finishing the test runs.

